PR #114, opened 2015-02-06
Synthesized events cause crashes on firefox

kmansel:


src/joint.dia.paper.js
```diff
@@ -620,6 +620,17 @@ joint.dia.Paper = Backbone.View.extend({
 
     snapToGrid: function(p) {
 
+        // Synthesized events such as in unit tests won't have clientX/clientY
```
kumilingus: Are you able to add some dummy `clientX` and `clientY` to synthesized events in your unit test?

PR #116, opened 2015-03-02
This update allows the ref-x, ref-y, ref-width, and ref-height to have t...

kmansel:
...he format "x".   Currently if the value you supply is between 0 and 1, then it will be a multiple of the width/height of the ref target, meaning we can only multiply by a number less than one.  If you want to make something's width twice the width of it's ref target, you can't.  This allows you to specify a number followed by "x" to force that number to be a multiplier, so for example ref-width: '3x' would make the element 3 times the width of the ref target.  This change is fully backward compatible.

### DIFF
--- a/src/joint.dia.element.js
+++ b/src/joint.dia.element.js
@@ -313,17 +313,17 @@ joint.dia.ElementView = joint.dia.CellView.extend({
 
         var ref = elAttrs['ref'];
         var refX = parseFloat(elAttrs['ref-x']);
-        var refXMult = _.isString(elAttrs['ref-x']) ? (elAttrs['ref-x'][elAttrs['ref-x']-1] === 'x') : false;
+        var refXMult = _.isString(elAttrs['ref-x']) ? (elAttrs['ref-x'][elAttrs['ref-x'].length-1] === 'x') : false;
         var refY = parseFloat(elAttrs['ref-y']);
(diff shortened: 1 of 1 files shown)


src/joint.dia.element.js
```diff
@@ -313,13 +313,17 @@ joint.dia.ElementView = joint.dia.CellView.extend({
 
         var ref = elAttrs['ref'];
         var refX = parseFloat(elAttrs['ref-x']);
+        var refXMult = _.isString(elAttrs['ref-x']) ? (elAttrs['ref-x'][elAttrs['ref-x']-1] === 'x') : false;
```
kumilingus: Hi Kevin, are you missing `length` here by any chance? i.e. `elAttrs['ref-x'][elAttrs['ref-x'].length-1]`
`elAttrs['ref-x'][elAttrs['ref-x']-1]` is always `undefined` (except for `elAttrs['ref-x'] = "1"`).

kmansel: Yes you're right!  Adding update right now.

kumilingus:
Hi Kevin,
we suggest to allow users specify the percentage for all ref's instead of multiplier (as in CSS). e.g.

    { 'ref': '.body', 'ref-x': '300%', 'ref-y': '50%' }
    // equivalent to '3x' and '0.5'

What do you reckon?

kumilingus:
Users will be able to specify all `ref` attributes as a percentage of the referenced value (e.g. 'ref-x': '300%') in very next JointJS release. Thanks for suggestion Kevin.

PR #118, opened 2015-03-11
Prevent errors while removing link when embedded shape is removed earlier

jefsnare:


DavidDurman:
Do you have an example when this happens? Just want to see how come that the parent cell ID is stored while the cell is not in the collection anymore.

chill117:
I think this issue can be closed.

jefsnare:
No this can't be closed and is an issue.
I don't have the time to create a testcase for now, but if you create two elements, one parent one child. and you remove the parent, an error occurs.

kumilingus:
Hi Jeffrey,
when you have some spare time please try to create a testcase. We are not able to reproduce, what you just described. No errors occur.
Also note that we are not able to merge your PR as it does not satisfy our code styling rules (Run `make cstest`) and as the code indentation has been fixed long time ago,  the PR has to be rebased with master branch first.

PR #119, opened 2015-03-12
Fixed issue with FF 36+

rarous:
Firefox can crash on rendering of tools when the length of links is not known yet.

### DIFF
--- a/src/joint.dia.link.js
+++ b/src/joint.dia.link.js
@@ -1042,7 +1042,7 @@ joint.dia.LinkView = joint.dia.CellView.extend({
 
     getPointAtLength: function(length) {
 
-        return this._V.connection.node.getPointAtLength(length);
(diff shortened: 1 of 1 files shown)


DavidDurman:
Do you have an example of when FF crashes on this? I think we should fix the core of the problem rather than patching it. It is assumed one calls `getPointAtLength()` with valid integer number. If that doesn't happen somewhere, it should be fixed there.

rarous:
I was using Dagre plugin. When I called `resetCells` before `layout` it crashed. It started after FF36 release, older releases were fine.

chill117:
I am not able to reproduce this problem. Could you provide a code sample which replicates the problem?

chill117:
Fixed by PR #136

kumilingus:
Thanks @rarous for pointing this out.

PR #132, opened 2015-04-20
Exclude ancestor elements from routing in order to account for routing be...

automagic:
...tween embedded elements

https://github.com/DavidDurman/joint/issues/131

### DIFF
--- a/plugins/routers/joint.routers.manhattan.js
+++ b/plugins/routers/joint.routers.manhattan.js
@@ -382,7 +382,31 @@ joint.routers.manhattan = (function() {
             .difference(excludedEnds)
             // remove all elements whose type is listed in excludedTypes array
             .reject(function(element) {
-                return _.contains(opt.excludeTypes, element.get('type'));
+                // reject any element which is an ancestor of either source or target
+                var excludeAncestors = [];
+
+                var findAncestors = function(cell, arr){
+                    if(cell === undefined)
+                        return;
+
+                    var parent = cell.get('parent');
+                    if(parent !== undefined){
+                        arr.push(parent);
+                        findAncestors(graph.getCell(parent), arr);
+                    }
+                };
+
+                var sourceId = link.get('source').id;
+                if(sourceId !== undefined) {
+                    findAncestors(graph.getCell(sourceId), excludeAncestors);
+                }
+
+                var targetId = link.get('target').id;
+                if(targetId !== undefined) {
+                    findAncestors(graph.getCell(targetId), excludeAncestors);
+                }
+
+                return _.contains(opt.excludeTypes, element.get('type')) || _.contains(excludeAncestors, element.id);
             })
             // change elements (models) to their bounding boxes
             .invoke('getBBox')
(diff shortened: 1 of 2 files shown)


plugins/routers/joint.routers.manhattan.js
```diff
@@ -382,7 +382,31 @@ joint.routers.manhattan = (function() {
             .difference(excludedEnds)
             // remove all elements whose type is listed in excludedTypes array
             .reject(function(element) {
-                return _.contains(opt.excludeTypes, element.get('type'));
+                // reject any element which is an ancestor of either source or target
+                var excludeAncestors = [];
```
kumilingus: It's not necessary to compute the `excludeAncestors` array for every single element. All ancestors can be found in outer scope (source and target of the link stays the same). Note that performance matters in this case.

plugins/routers/joint.routers.manhattan.js
```diff
@@ -382,7 +382,31 @@ joint.routers.manhattan = (function() {
...
+
+                var findAncestors = function(cell, arr){
```
kumilingus: Would be nice if we could have `getAncestors()` method on the `joint.dia.Cell` prototype itself. Returning an array of ancestor cells (or `[]` if the cell has no parent).

PS: I would rather use an iteration (`while (parent) { .. }`) instead of a recursion.

kumilingus:
Thanks James for this PR! The issue made 'manhattan' routing almost unusable with embedded elements.
Please let me know if you are going to implement the `getAncestors()` method I proposed. I would merge it as it is and do it myself later on otherwise. Cheers!

plugins/routers/joint.routers.manhattan.js
automagic: Yes good point, I will move it

plugins/routers/joint.routers.manhattan.js
automagic: Will do.

src/joint.dia.cell.js
```diff
@@ -261,6 +261,27 @@ joint.dia.Cell = Backbone.Model.extend({
...
+    getAncestors: function(){
+
+        var ancestors = [],
+            parentId = this.get('parent');
+
+        if(this.collection === undefined)
+            return ancestors;
+
+        while(parentId !== undefined) {
+            var parent = _.find(this.collection.models, function(item){ return item.id === parentId; });
```
kumilingus: `this.collection` is a Backbone collection. So we can call directly `this.collection.get(parentId)` here.

kumilingus:
One more comment about code styling :) Can you run `jscs src/joint.dia.cell.js` and `jscs plugins/routers/joint.routers.manhattan.js', fix the code styling issues and push the changes.
You can even run `jscs -x filename` to automatically fix issues. It should handle most of them. Other needs to be fixed manually.
Sorry, I do realize that it is quite a new thing and it has not been documented yet.

automagic:
You mean to say my code isn't beautiful? ;)

kumilingus:
No way! It's the most beautiful code I've seen for a long long time! :) But..
It's different.
It's different from the rest of the library.

kumilingus:
N1 James! Thanks.

kumilingus:
A unit test would be also nice. But who can ask for that, right? ;)

kumilingus:
A test is there. I've also changed that line to find a parent cell. Thx once again.

PR #133, opened 2015-04-21
Fix issue with linkView poinermove and pointerup events being called before work is done.

automagic:
Fixes batch:stop being called before the work in the batch is complete (in this case assigning link target on pointerup).

Makes link and element pointer events more uniform,  and corrects calls to batch:start and batch:stop

### DIFF
--- a/src/joint.dia.element.js
+++ b/src/joint.dia.element.js
@@ -694,11 +694,11 @@ joint.dia.ElementView = joint.dia.CellView.extend({
 
     pointerdown: function(evt, x, y) {
 
-        this.model.trigger('batch:start');
-
         // target is a valid magnet start linking
         if (evt.target.getAttribute('magnet') && this.paper.options.validateMagnet.call(this.paper, this, evt.target)) {
 
+            this.model.trigger('batch:start');
+            
             var link = this.paper.getDefaultLink(this, evt.target);
             link.set({
                 source: {
@@ -712,14 +712,15 @@ joint.dia.ElementView = joint.dia.CellView.extend({
             this.paper.model.addCell(link);
 
             this._linkView = this.paper.findViewByModel(link);
+            this._linkView.pointerdown(evt, x, y);
             this._linkView.startArrowheadMove('target');
 
         } else {
-
             this._dx = x;
             this._dy = y;
 
             joint.dia.CellView.prototype.pointerdown.apply(this, arguments);
+            this.notify('element:pointerdown', evt, x, y);
         }
     },
 
@@ -765,7 +766,9 @@ joint.dia.ElementView = joint.dia.CellView.extend({
             this._dx = g.snapToGrid(x, grid);
             this._dy = g.snapToGrid(y, grid);
 
+
             joint.dia.CellView.prototype.pointermove.apply(this, arguments);
+            this.notify('element:pointermove', evt, x, y);
         }
     },
 
@@ -775,8 +778,9 @@ joint.dia.ElementView = joint.dia.CellView.extend({
 
             // let the linkview deal with this event
             this._linkView.pointerup(evt, x, y);
-
             delete this._linkView;
+
+            this.model.trigger('batch:stop');
 
         } else {
 
@@ -785,10 +789,10 @@ joint.dia.ElementView = joint.dia.CellView.extend({
                 this._inProcessOfEmbedding = false;
             }
 
+            this.notify('element:pointerup', evt, x, y);
             joint.dia.CellView.prototype.pointerup.apply(this, arguments);
-        }
 
-        this.model.trigger('batch:stop');
+        }
     }
 
 });
--- a/src/joint.dia.link.js
+++ b/src/joint.dia.link.js
@@ -1105,8 +1105,6 @@ joint.dia.LinkView = joint.dia.CellView.extend({
 
     _beforeArrowheadMove: function() {
 
-        this.model.trigger('batch:start');
-
         this._z = this.model.get('z');
         this.model.toFront();
 
@@ -1134,8 +1132,6 @@ joint.dia.LinkView = joint.dia.CellView.extend({
         if (this.paper.options.markAvailable) {
             this._unmarkAvailableMagnets();
         }
-
-        this.model.trigger('batch:stop');
     },
 
     _createValidateConnectionArgs: function(arrowhead) {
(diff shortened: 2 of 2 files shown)


kumilingus:
Hi James,
I like the idea. But I'm getting some unpleasant results in some cases.

Please open `demo/basic.html` and the web inspector. Run the following code.

    graph.on('batch:start', function() { console.log('start'); });
    graph.on('batch:stop', function() { console.log('stop'); });

And now
1) Create a new link by drag and drop a port. You can see `start, start, stop` printed in the console. And also there is an unwanted vertex created just next to the port.
2) Remove a link by clicking the remove button on the link. You can see `start, start, stop` printed in the console (means that there is one pending batch).

automagic:
To me the batch events look correct, i am getting a start on link.pointerdown, a start/stop on model.toFront and a stop on link.pointerup.  However the extra vertex is definitely an issue, i need to fix.

automagic:
Ok, it should not be creating the extra vertex.   The batch start and stop are paired correctly... there is one less than before, as I removed the arrowhead start / stop on 1108.  I guess I am not sure what the nested batches are for, so I can add it back.

try adding: 
`paper.on('link:pointerdown', function() { console.log('pointerdown'); });`
`paper.on('link:pointerup', function() { console.log('pointerup'); });`

in addition to the batch start/stop logging and let me know what you think.

src/joint.dia.link.js
```diff
@@ -1230,89 +1226,89 @@ joint.dia.LinkView = joint.dia.CellView.extend({
...
-        var parentClassName = evt.target.parentNode.getAttribute('class');
-        var labelNode;
-        if (parentClassName === 'label') {
-            className = parentClassName;
-            labelNode = evt.target.parentNode;
-        } else {
-            labelNode = evt.target;
-        }
+        // if are simulating pointerdown on a link during a magnet click, skip link interactions
+        if (evt.target.getAttribute('magnet') == null) {
```
kumilingus: Can we change this to `if (evt.target.getAttribute('magnet') != null) return;` to prevent unnecessary indenting?

src/joint.dia.element.js
```diff
@@ -694,8 +694,6 @@ joint.dia.ElementView = joint.dia.CellView.extend({
 
     pointerdown: function(evt, x, y) {
 
-        this.model.trigger('batch:start');
-
         // target is a valid magnet start linking
         if (evt.target.getAttribute('magnet') && this.paper.options.validateMagnet.call(this.paper, this, evt.target)) {
```
kumilingus: Here we are missing a batch now. A new link is added to the graph with no previous batch start resulting that one user action (from `pointerdown` to `pointerup`) it's not entirely wrapped in a batch.
I'm not sure what are you using batches for, but one of the usage might be a history (undo/redo) implementation, where creating a link, dragging its arrowhead and then connecting it to an element should be an atomic action.

src/joint.dia.element.js
```diff
@@ -775,7 +776,6 @@ joint.dia.ElementView = joint.dia.CellView.extend({
 
             // let the linkview deal with this event
             this._linkView.pointerup(evt, x, y);
-
             delete this._linkView;
```
kumilingus: and `batch:stop` here. Once I merge this PR I'm going to add names for the batches. e.g.

    this.model.trigger('batch:stop', { batchName: 'add-link' });

which might be useful not only for debugging purposes.

src/joint.dia.cell.js
```diff
@@ -980,6 +980,7 @@ joint.dia.CellView = Backbone.View.extend({
     pointermove: function(evt, x, y) {
 
         this.notify('cell:pointermove', evt, x, y);
+
```
kumilingus: If `joint.dia.cell.js` doesn't have to be changed after all. Can you please remove this extra lines?
And once everything else is finished and ready for merging to the upstream can you merge all your commits into a one so we keep the history clearer and better to read?

    git rebase -i HEAD~[number-of-commits]

src/joint.dia.element.js
automagic: The batch:start is getting called in CellView.prototype.pointerdown, duing both Element:pointerdown and Link:pointerown; adding it back here causes a extra unnecessary bactch:start;  For adding a link it becomes Element:batch:start- Link:batch:start; but since you are really just passing the pointerdown to the link handers on magnet click, it seems unnecessary;

Since the batch:start/stop is necessary in the CellView prototype due to having to deal with those events on delete, in my opinion they are not necessary in the derived.  In this implmentation we are relying on the CellView prototype to handle the batch start/stop, which is why they arent present in LinkView or CellView pointer down/up

src/joint.dia.element.js
automagic: If this is the approach you want to take, then it makes more sense to add it back in (since they are now named), but in my implementation, batch start is called once on LinkView models' pointerdown (via its protoype) and same for ElementView.

src/joint.dia.element.js
```diff
@@ -775,7 +776,6 @@ joint.dia.ElementView = joint.dia.CellView.extend({
 
             // let the linkview deal with this event
             this._linkView.pointerup(evt, x, y);
```
automagic: batch stop called here

src/joint.dia.element.js
kumilingus: I see, but I don't think it is unnecessary. The problem is that `batch:start` is triggered way to late now. Suppose you are listening to graph on `all` events and you create a link by dragging & dropping a port.

`add`, `batch:start`, `change:target`, `change:target`, `change:target`, `batch:stop` is what you see.

`batch:start`, `add`, `batch:start`, `change:target`, `change:target`, `change:target`, `batch:stop`, `batch:stop` is what you want to see (or alternatively `batch:start`, `add`, `change:target`, `change:target`, `change:target`, `batch:stop`, but an extra batch is not an issue here).

If you add some batch names it starts making more sense I think.
- cell batch:start/stop :  `{ batchName: 'pointer' }`
- element (a valid magnet branch) batch:start/stop : `{ batchName: 'add-link' }`

`batch:start(add-link)`, `add`, `batch:start(pointer)`, `change:target`, `change:target`, `change:target`, `batch:stop(pointer)`, `batch:stop(add-link)`

automagic: The problem I am trying to solve with the batch:start and batch:stop is the double call, once in the derived class  and once in the base class (ElementView.pointerdown   / CellView.pointerdown).   Since the base class call is essentially inlined (by applying this) the event is triggered twice on the same model- e.g. this.model.trigger('batch:start') where this.model is Element.   Since we are always invoking the base class event, it seems correct to just handle it there.

automagic: I do see your point about the 'add' happening before the batch, however.. I will add it back in

kumilingus: James I think you've kind of done what you described. I can not see any double calls anymore.
- A batch named `pointer` can be triggered by a link or element. That's why it's handled on the cellView.
- A batch names `add-link` can be triggered by an element only. That's why it's handled on the ElementView.
- Note also there is a bunch of other `batch:start`/`batch:stop` triggered on the model (`to-front`, `to-back`, `embed`, `unembed`, `resize`, `remove`, `clear`). 

I'm very happy with what you've done here. I think it's ready now. Do you mind run `git rebase -i HEAD~10` and squash commits before we merge it? Cheers.

automagic:
Commits squashed and rebased

kumilingus:
Thanks!

PR #137, opened 2015-05-07
Added the possibility of multiple extensions on the Ports interfaces

Rodsevich:
I was extending the functionality of the Elements, but when using Ports, their's interfaces were made to call their inmediate parent function's, forbidding the option of having several interfaces implemented unless the port's interface were the last used to extend the functionality.
Just changed that code for having better options when implementing interfaces


kumilingus:
Fixed in 14c96c184f3f52adb1980e5caf012e695136557e (ports reimplemented).

PR #140, opened 2015-05-11
Add "main" section to bower.json to support build processes

ytechie:
I added a "main" section so that things like wiredep can automatically wire-up the script/css tags in HTML to include the resources.


kumilingus:
Great! Thank you!

PR #144, opened 2015-06-04
if an element in the list is positioned relative to this one, then we…

kmansel:
… want to insert this one before it in the list


kumilingus:
Thank you Kevin for pointing that out. The fix is now implemented in #502.

PR #145, opened 2015-06-07
Allow ports with arbitrary markup

imuli:
have getPortSelector return a selector to .port-body instead of circle.

This is compatable with the DEVS extension, but may break third party
code if they didn't copy over 'class="port-body" in their portMarkup.

Thoughts?


kumilingus:
We'll announce a breaking change in the next release. Hope not many developers will get affected (you are right - only those using custom markups for ports with no `port-body` class). This PR is in line with original idea to select ports by a class name and not by specific tag. Thx!

PR #146, opened 2015-06-10
fix ref-x and ref-y to match behavior comment

imuli:
Alternately fix would be to update the comment. Trade off between being able to set an object just below / right of its reference or 1 pixel from the top/left.


chill117:
Thanks for the effort here. We corrected the documentation to match the actual behavior, rather than change the behavior. You can see the change [here](https://github.com/clientIO/joint/pull/315).

PR #153, opened 2015-07-07
CommonJS / AMD Issues

rfastturnlow:
I pulled down the latest to try out the new module support and ran into a few minor issues with WebPack:

1. The library paths in the grunt config file have not been updated since switching to bower.  This isn't really an issue with module support, but just something I noticed when trying to decide which output file to use.
2. The two new concat tasks for the build script output files were missing parts of their template strings because grunt was trying to substitute in replacement values during the concat task.  I've updated the tasks to use the 'square' delimiter only and the template strings made it it to the output intact.
3. The vectorizer module was not properly importing the geometry library.  I'm not quite sure why vectorizer and geometry weren't named for AMD, but I went ahead and gave them default names to match their global names and then added g as an import to vectorizer.  Doing so meant that g is coming from the closure now so the geometry file must be included before the vectorizer file in most situations.  To remedy that I swapped the file order in the concat tasks and any html file that included the files through script tags.

  Lastly I added a sad test to demonstrate the issue in the requirejs test file.  I call it sad because the only way I could think to demonstrate the issue was to access a method in vectorizer that depended on the geometry library.  Obviously implementation details could change that will invalidate that so its long term value is pretty low.

That's it.  A couple minor fixes and the modules are working great.  Thanks for all your hard work on a great library!


kumilingus:
Hi Randy,

Note that `vectorizer` supposed to have no `geometry` dependency at all. Your changes and the test is a perfect solution before making `vectorizer` standalone takes a place.

Thumb up for all changes you've made. Thank you.

rfastturnlow:
Just as a follow up for the sake of other WebPack users.  The imports-loader would have worked to inject that dependency as well.  I ended up using that to include the DirectedGraph layout module on its own to the clean build bundle using this include:

    require('imports?joint&_=lodash!jointjs/dist/joint.layout.DirectedGraph.js');

PR #155, opened 2015-07-12
Support optional custom markup for second link tools

jbaayen:


kumilingus:
Seems useful. Thank You!

PR #164, opened 2015-08-01
Fix defaulting in `Element.renderMarkup()`

stackfull:
The code doesn't match the comment. It should try to get "markup" from the model first and fallback to the prototype.


kumilingus:
Well, that makes sense. Thank You!

PR #167, opened 2015-08-03
Compiled lodash templates

automagic:
The use of the lodash function _.template violates modern browser Content Security Policy (CSP) 'unsafe-eval'.   See https://github.com/lodash/lodash/issues/832

This pull request removes _.template calls in favor of precompiling the templates using a grunt task and including them in the build.  

We wanted to float this idea and get some feedback


mike-goodwin:
Is there any plan to merge this into master? It would be nice to remove unsafe-eval...

chill117:
There might be more changes needed to make jointjs work well with CSP. I'll add this to my to-do list.

chill117:
@mike-goodwin I've updated jointjs with fixes that will allow you to use the content-security-policy header  without unsafe-eval. These changes will be included in the next jointjs release (0.9.8). Here's the CSP header that will work with jointjs:
```
add_header Content-Security-Policy "default-src 'self'; img-src 'self' data:;";
```

mike-goodwin:
@chill117 : Thank you - I'll upgrade to 0.9.8 as soon as it is released!

PR #172, opened 2015-08-24
Fixed bug where link z order is corrupt when previous state was 0 on head move

rickbutton:
If a link has a Z order of 0, then after an arrowhead move the original z order will
be corrupted (set to Number.MAX_VALUE instead of back to 0). This was due to incorrect
logic [if(0), 0 evaluates to false in this case]. Replacing this with an undefined check
is the correct behavior to handle a 0 z order value.


kumilingus:
Well spotted. Thank You.

PR #175, opened 2015-08-30
added type for pn.Link (fixed #174)

### LINKED ISSUE #174: Type of pn.Link is not serialized correctly.
[joint.shapes.pn.js](https://github.com/clientIO/joint/blob/master/plugins/joint.shapes.pn.js) defines pn.Link in joint.shapes. When the graph is serialized into json, the type of the connection is not `pn.Link`, but `link`. When deserialized the paper shows now normal links instead of pn.Link.

SrTobi:


kumilingus:
Thank You!

PR #222, opened 2016-01-31
Geometry: add scale(), toJSON() methods

kumilingus:
dia.Element: add scale() method
dia.Link: add scale() method
dia.Graph: add resizeCells() method

### DIFF
--- a/src/joint.dia.link.js
+++ b/src/joint.dia.link.js
@@ -93,30 +93,46 @@ joint.dia.Link = joint.dia.Cell.extend({
 
     translate: function(tx, ty, opt) {
 
-        var attrs = {};
-        var source = this.get('source');
-        var target = this.get('target');
-        var vertices = this.get('vertices');
-
-        if (!source.id) {
-            attrs.source = { x: (source.x || 0) + tx, y: (source.y || 0) + ty };
-        }
-
-        if (!target.id) {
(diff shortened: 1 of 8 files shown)


src/joint.dia.link.js
```diff
@@ -93,31 +93,43 @@ joint.dia.Link = joint.dia.Cell.extend({
...
+    },
+
+    scale: function(sx, sy, origin, opt) {
+
+        return this.applyOnPoints(function(p) {
+            return g.point(p).scale(sx, sy, origin).toJSON();
+        }, opt);
+    },
+
+    applyOnPoints: function(fn, opt) {
```
chill117: I like this. But I would like tests too :)

test/jointjs/graph.js
```diff
@@ -114,640 +114,705 @@ module('graph', {
...
-        },
-        embeddedLink1: {
-            source: embeddedLink1.get('source'),
-            target: embeddedLink1.get('target')
-        }
-    };
-
-    assert.equal(after.rect1.bbox.x, before.rect1.bbox.x + dx, 'position/x of element');
-    assert.equal(after.rect1.bbox.y, before.rect1.bbox.y + dy, 'position/y of element');
-    assert.equal(after.l
```
chill117: For consistency, I think this should be a separate method called `graph.resize()` which resizes all the cells in the graph. `graph.resizeCells` would behave similarly to `graph.addCells`.

PR #229, opened 2016-02-02
Added 'cell:mousewheel' and 'blank:mousewheel' events to paper

chill117:


src/joint.dia.cell.js
```diff
@@ -942,6 +942,11 @@ joint.dia.CellView = joint.mvc.View.extend({
         this.notify('cell:mouseout', evt);
     },
 
+    mousewheel: function(evt, x, y) {
```
kumilingus: `delta` argument missing?

PR #241, opened 2016-02-10
Added support for y-alignment bottom and x-alignment right 

B3rn475:
In this way it is possible to align an element w.r.t. the bottom or right of the parent.
Right now it is not possible and is required to use an y-alignment or x-alignment of  -0.9999999


kumilingus:
That's handy. Thanks.

PR #243, opened 2016-02-10
Avoid to call sort too much

B3rn475:
Call sort just one time during toFront or toBack deep operations see issue #249

### DIFF
--- a/src/joint.dia.graph.js
+++ b/src/joint.dia.graph.js
@@ -17,7 +17,6 @@ joint.dia.GraphCells = Backbone.Collection.extend({
                 --activeBatches;
             }
         }, this);
-    
         // Backbone automatically doesn't trigger re-sort if models attributes are changed later when
         // they're already in the collection. Therefore, we're triggering sort manually here.
         this.on('change:z', function() { return activeBatches === 0 && this.sort(); }, this);
--- a/src/joint.dia.graph.js
+++ b/src/joint.dia.graph.js
@@ -7,9 +7,20 @@ joint.dia.GraphCells = Backbone.Collection.extend({
 
     initialize: function(models, opt) {
 
+        //In order to avoid multiple sort during toFront and toBack deep operations
+        //The change:z event is disabled during batch:start and reactivated when all the batches have stopped
+        var activeBatches = 0;
+        this.on('batch:start', function (e) { if ((e.batchName === 'to-front') || (e.batchName === 'to-back')) { ++activeBatches; }});
+        this.on('batch:stop', function (e) {
+            if (activeBatches > 0 && ((e.batchName === 'to-front') || (e.batchName === 'to-back'))) {
+                this.sort();
+                --activeBatches;
+            }
+        }, this);
+    
         // Backbone automatically doesn't trigger re-sort if models attributes are changed later when
         // they're already in the collection. Therefore, we're triggering sort manually here.
-        this.on('change:z', this.sort, this);
+        this.on('change:z', function () { return activeBatches === 0 && this.sort(); }, this);
 
         // Set the optional namespace where all model classes are defined.
         if (opt.cellNamespace) {
--- a/src/joint.dia.graph.js
+++ b/src/joint.dia.graph.js
@@ -10,8 +10,8 @@ joint.dia.GraphCells = Backbone.Collection.extend({
         //In order to avoid multiple sort during toFront and toBack deep operations
         //The change:z event is disabled during batch:start and reactivated when all the batches have stopped
         var activeBatches = 0;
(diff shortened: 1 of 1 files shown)


B3rn475:
I've made this Pull Request because i often call toFront({deep: true}) on Elements with tens of descendant.
In this situation every time a z changes all the elements in the paper are sorted.
Given n (the number of descendants) and m (the number of operations required to sort the paper).
Right now the complexity of a toFront({deep: true}) is O(n * m).
After my changes it becomes O(n + m).

chill117:
Looks good. I've tested it with both JointJS and Rappid, and everything is still working.

I will merge it now, but with the plan of refactoring what you've done into a proper API on the graph model so that we can replace the following:
```js
        //In order to avoid multiple sort during toFront and toBack deep operations
        //The change:z event is disabled during batch:start and reactivated when all the batches have stopped
        var activeBatches = 0;
        this.on('batch:start', function(e) { if ((e.batchName === 'to-front') || (e.batchName === 'to-back')) { ++activeBatches; }});
        this.on('batch:stop', function(e) {
            if (activeBatches > 0 && ((e.batchName === 'to-front') || (e.batchName === 'to-back'))) {
                this.sort();
                --activeBatches;
            }
        }, this);
        // Backbone automatically doesn't trigger re-sort if models attributes are changed later when
        // they're already in the collection. Therefore, we're triggering sort manually here.
        this.on('change:z', function() { return activeBatches === 0 && this.sort(); }, this);
```
with:
```js
        // Backbone automatically doesn't trigger re-sort if models attributes are changed later when
        // they're already in the collection. Therefore, we're triggering sort manually here.
        this.on('change:z', function() { return !this.hasActiveBatch() && this.sort(); }, this);
```

B3rn475:
No problem for the refactoring, but watch out that it is not the same.
If you do that it will not sort anymore after a toFront or a toBack.
Due to the toFront implementation no change:z will be triggered when there are not ActiveBatches.
I think that the right solution would be
```javascript
        // Backbone automatically doesn't trigger re-sort if models attributes are changed later when
        // they're already in the collection. Therefore, we're triggering sort manually here.
        var hasToBeSorted = false;
        this.on('change:z', function() {
            if(!this.hasActiveBatch()) {
                hasToBeSorted = false;
                this.sort();
            } else {
                hasToBeSorted = true;
            }
        }, this);
        this.on('batch:stop', function() { return hasToBeSorted && this.sort(); }, this);
```

chill117:
How about this?
```js
this.on('change:z', function() { return !this.hasActiveBatch('to-front') && !this.hasActiveBatch('to-back') && this.sort(); }, this);
```
The `hasActiveBatch` method will accept an optional `name` (batch name) argument.

B3rn475:
The problem is the following.
The implementation of toFront is
```javascript
this.trigger('batch:start', {batchName: 'to-front'});
// change the z of the elements even recursively
this.trigger('batch:stop, {batchName: 'to-front'});
```
In this situation a 'change:z' will be triggered only during a batch and so no sorting will be triggered anymore.
```javascript
this.on('change:z batch:stop', function() { return !this.hasActiveBatch('to-front') && !this.hasActiveBatch('to-back') && this.sort(); }, this);
```
But you will still sort after a batch that is not of tipe to-front or to-back.
To avoid this a solution can be
```javascript
 // Backbone automatically doesn't trigger re-sort if models attributes are changed later when
 // they're already in the collection. Therefore, we're triggering sort manually here.
 this.on('change:z batch:stop', function() {
       if (e.batchName !== 'to-front' && e.batchName !== 'to-back') { return; }
       return !this.hasActiveBatch('to-front') && !this.hasActiveBatch('to-back') && this.sort();
 }, this);
```
I don't want to be a pain in the neck, but this performance issue is what makes my project unusable in many situations.

chill117:
> I don't want to be a pain in the neck, but this performance issue is what makes my project unusable in many situations.

You are not being a pain in the neck. It is important to hash out things clearly so that we implement a solid solution. We've already merged your changes, so the performance improvement will be there for the next release. Now I am just trying to figure out a way to make doing things with batches (like you needed to do) a bit easier / clearer.

B3rn475:
Mixing the solutions up we can do this.
```javascript
this.on('change:z', function() { return !this.hasActiveBatch('to-front') && !this.hasActiveBatch('to-back') && this.sort(); }, this);
this.on('batch:stop', function() { return (e.batchName === 'to-front' || e.batchName === 'to-back') && !this.hasActiveBatch(e.batchName) && this.sort(); }, this);
```
What do you think about?

chill117:
Yes, that is a bit cleaner. I think we will want to add some unit tests for these cases when the code is refactored. I'll link to this thread from the issue I mentioned.

Issue #239, opened 2016-02-09
Batch-related API improvements for graph model


chill117:
Currently start/stopping batches of graph changes looks like this:
```js
graph.trigger('batch:start', { batchName: 'some_batch', other: 'info' });
// ..
graph.trigger('batch:stop', { batchName: 'some_batch', other: 'info' });
```

A cleaner looking approach could be:
```js
graph.startBatch('layout', { other: 'info' });
// ..
graph.stopBatch('layout', { other: 'info' });
```
Where..
* the first argument (batch name) is required
* the second argument (batch data) is optional

Under-the-hood, the old way of doing batches (triggering the `batch:start` and `batch:stop` events) could be used. This would be a backwards compatible change.

Also we should add the following method for detecting if there is a batch in progress:
```js
graph.hasActiveBatch([name])
```
Where..
* the `name` argument is optional

kumilingus:
I like that. I would call methods `startBatch()` and `stopBatch()` though (verbs).

chill117:
See [the comment thread on PR #243](https://github.com/clientIO/joint/pull/243) for places that could benefit from these new batch methods.

PR #250, opened 2016-02-15
Implemented enhancement #239 with unit test for to-front and to-back  

B3rn475:
See issue #239

### DIFF
--- a/src/joint.dia.cell.js
+++ b/src/joint.dia.cell.js
@@ -149,7 +149,7 @@ joint.dia.Cell = Backbone.Model.extend({
         // Store the graph in a variable because `this.graph` won't' be accessbile after `this.trigger('remove', ...)` down below.
         var graph = this.graph;
         if (graph) {
-            graph.trigger('batch:start', { batchName: 'remove' });
+            graph.startBatch('remove');
         }
 
         // First, unembed this cell from its parent cell if there is one.
@@ -165,7 +165,7 @@ joint.dia.Cell = Backbone.Model.extend({
         this.trigger('remove', this, this.collection, opt);
 
         if (graph) {
-            graph.trigger('batch:stop', { batchName: 'remove' });
+            graph.stopBatch('remove');
         }
 
         return this;
@@ -179,7 +179,7 @@ joint.dia.Cell = Backbone.Model.extend({
 
             var z = (this.graph.getLastCell().get('z') || 0) + 1;
 
-            this.trigger('batch:start', { batchName: 'to-front' }).set('z', z, opt);
+            this.startBatch('to-front').set('z', z, opt);
 
             if (opt.deep) {
 
@@ -188,7 +188,7 @@ joint.dia.Cell = Backbone.Model.extend({
 
             }
 
-            this.trigger('batch:stop', { batchName: 'to-front' });
+            this.stopBatch('to-front');
         }
 
         return this;
@@ -202,7 +202,7 @@ joint.dia.Cell = Backbone.Model.extend({
 
             var z = (this.graph.getFirstCell().get('z') || 0) - 1;
 
-            this.trigger('batch:start', { batchName: 'to-back' });
+            this.startBatch('to-back');
 
             if (opt.deep) {
 
@@ -210,7 +210,7 @@ joint.dia.Cell = Backbone.Model.extend({
                 _.eachRight(cells, function(cell) { cell.set('z', z--, opt); });
             }
 
-            this.set('z', z, opt).trigger('batch:stop', { batchName: 'to-back' });
+            this.set('z', z, opt).stopBatch('to-back');
         }
 
         return this;
@@ -224,7 +224,7 @@ joint.dia.Cell = Backbone.Model.extend({
 
         } else {
 
-            this.trigger('batch:start', { batchName: 'embed' });
+            this.startBatch('embed');
 
             var embeds = _.clone(this.get('embeds') || []);
 
@@ -234,7 +234,7 @@ joint.dia.Cell = Backbone.Model.extend({
             cell.set('parent', this.id, opt);
             this.set('embeds', _.uniq(embeds), opt);
 
-            this.trigger('batch:stop', { batchName: 'embed' });
+            this.stopBatch('embed');
         }
 
         return this;
@@ -242,12 +242,12 @@ joint.dia.Cell = Backbone.Model.extend({
 
     unembed: function(cell, opt) {
 
-        this.trigger('batch:start', { batchName: 'unembed' });
+        this.startBatch('unembed');
 
         cell.unset('parent', opt);
         this.set('embeds', _.without(this.get('embeds'), cell.id), opt);
 
-        this.trigger('batch:stop', { batchName: 'unembed' });
+        this.stopBatch('unembed');
 
         return this;
     },
@@ -608,6 +608,16 @@ joint.dia.Cell = Backbone.Model.extend({
     isLink: function() {
 
         return false;
+    },
+
+    startBatch: function(name, opt) {
+        if (this.graph) { this.graph.startBatch(name, _.extend({}, opt, { cell: this })); }
+        return this;
+    },
+
+    stopBatch: function(name, opt) {
+        if (this.graph) { this.graph.stopBatch(name, _.extend({}, opt, { cell: this })); }
+        return this;
     }
 });
 
@@ -907,9 +917,9 @@ joint.dia.CellView = joint.mvc.View.extend({
 
     pointerdown: function(evt, x, y) {
 
-        if (this.model.collection) {
-            this.model.trigger('batch:start', { batchName: 'pointer' });
-            this._collection = this.model.collection;
+        if (this.model.graph) {
+            this.model.startBatch('pointer');
+            this._graph = this.model.graph;
         }
 
         this.notify('cell:pointerdown', evt, x, y);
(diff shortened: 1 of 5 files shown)


src/joint.dia.graph.js
```diff
@@ -51,6 +51,22 @@ joint.dia.GraphCells = Backbone.Collection.extend({
     comparator: function(model) {
 
         return model.get('z') || 0;
+    },
+
+    startBatch: function(name, opt) {
+        return this.trigger('batch:start', _.extend({}, opt, { element: this, batchName: name }));
```
kumilingus: Note that `this` context is not a `joint.dia.Element` but `joint.dia.Graph` here.
Batches in general doesn't have to be related to a single element. It could be related to a set of elements and links (e.g subgraph removal).
I suggest not to mention the `element` attribute here at all.

src/joint.dia.cell.js
```diff
@@ -608,6 +608,14 @@ joint.dia.Cell = Backbone.Model.extend({
     isLink: function() {
 
         return false;
+    },
+
+    startBatch: function(name, opt) {
+        return this.trigger('batch:start', _.extend({}, opt, { element: this, batchName: name }));
```
kumilingus: Batches makes sense only if an element is part of a graph. I would take advantage of the fact, that `startBatch` is already defined on the graph. i.e.
```javascript
if (this.graph) {
  return this.graph.startBatch(name, _.extend({ element: this }, opt));
}
```

B3rn475:
That was a "copy and paste" error.
Sorry

src/joint.dia.graph.js
```diff
@@ -5,21 +5,21 @@ joint.dia.GraphCells = Backbone.Collection.extend({
...
-        //The change:z event is disabled during batch:start and reactivated when all the batches have stopped
-        var activeBatches = 0;
-        this.on('batch:start', function(e) { if ((e.batchName === 'to-front') || (e.batchName === 'to-back')) { ++activeBatches; }});
-        this.on('batch:stop', function(e) {
-            if (activeBatches > 0 && ((e.batchName === 'to-front') || (e.batchName === 'to-back'))) {
-                this.sort();
-                --activeBatches;
-            }
-        }, this);
+        this.on('batch:start', function(e) { this.activeBatches[e.batchName] = (this.activeBatches[e.batchName] || 0) + 1; }, this);
```
kumilingus: Can we fallback to a default batch name if no `batchName` is provided? i.e
```javascript
initialize: function() {
   // ...
   this.on('batch:start', this.onBatchStart, this);
   // ..
},

onBatchStart: function(opt) {
  opt = opt || {};
  var batchName = opt.batchName || `default`;
  this.activeBatches[batchName] = (this.activeBatches[batchName] || 0) + 1;
}
```

src/joint.dia.cell.js
```diff
@@ -927,7 +935,7 @@ joint.dia.CellView = joint.mvc.View.extend({
         if (this._collection) {
             // we don't want to trigger event on model as model doesn't
             // need to be member of collection anymore (remove)
-            this._collection.trigger('batch:stop', { batchName: 'pointer' });
+            this._collection.stopBatch('pointer');
```
kumilingus: There is no `stopBatch` method defined on `collection` now. It's a legacy code anyway and this should be changed in the following way.
```javascript
if (this.model.graph) {
  this.model.startBatch('pointer');
  this._graph = this.model.graph;
}
  		  
if (this._graph) {
  // we don't want to trigger event on model as model doesn't
  // need to be member of graph anymore (remove)
  this._graph.startBatch('pointer');
  this._graph = null;
}
```

src/joint.dia.graph.js
B3rn475: Without setting a default, 'undefined' would be the default.
With this modification not only undefined but anything that is convertible to false like ``null``, ``0`` and ``''`` would be reset to default, but not ``1`` for example.
What is the expected behavior?

src/joint.dia.cell.js
B3rn475: GraphCells have stopBatch, should we move the code managing the Batches to the Graph instead?

src/joint.dia.graph.js
B3rn475: That was a "copy and paste" error.
Sorry

src/joint.dia.cell.js
B3rn475: Implemented.
Even though with a modification

```javascript
    startBatch: function(name, opt) {
        if (this.graph) { this.graph.startBatch(name, opt); }
        return this;
    }
```
The return is important because we would return the Graph and not the Cell instead, breaking chaining operations.

src/joint.dia.graph.js
```diff
@@ -75,6 +63,17 @@ joint.dia.Graph = Backbone.Model.extend({
         // to the outside world.
         cells.on('all', this.trigger, this);
 
+        this.on('batch:start', this._batchStarted, this);
+        this.on('batch:stop', this._batchStopped, this);
+        // Backbone automatically doesn't trigger re-sort if models attributes are changed later when
+        // they're already in the collection. Therefore, we're triggering sort manually here.
+        this.on('change:z', function(e) {
+            return !this.hasActiveBatch('to-front') && !this.hasActiveBatch('to-back') && cells.sort();
```
kumilingus: Handlers does not have to return a value + I would rather not to use constructs like `a && b()` but `if (a) { b() }` instead
```javascript
if (!this.hasActiveBatch('to-front') && !this.hasActiveBatch('to-back')) {
  cells.sort();
}
```

src/joint.dia.cell.js
kumilingus: Great, well spotted.

src/joint.dia.graph.js
B3rn475: Fixed

src/joint.dia.graph.js
kumilingus: With your approach `batchName` with `undefined` value and `null` falls into different categories.
`{ 'undefined':  1, ' null': 1 }`. But the meaning is kind of the same.

How about this?
```
var batchName = opt.batchName != null ? opt.batchName : 'undefined';
```

kumilingus:
Carlo, thanks for your time. It's a really great help.

B3rn475:
Thank you too.
Your amazing library saved me days of work.
I think is right to give some time back. ;-)

src/joint.dia.element.js
```diff
@@ -127,7 +127,7 @@ joint.dia.Element = joint.dia.Cell.extend({
 
         opt = opt || {};
 
-        this.trigger('batch:start', _.extend({}, opt, { element: this, batchName: 'resize' }));
+        this.startBatch('resize', _.extend({}, opt, { element: this }));
```
kumilingus: Maybe we should get rid of this `element` attribute equals `this` here.
The `joint.dia.Cell.prototype.startBatch` can add attribute `cell` equals `this` to the option object for all calls.

kumilingus:
I'm glad to hear that!

Btw I'd like to pull this PR first and test it manually before merging. For that I'll try find some time tomorrow.

src/joint.dia.cell.js
```diff
@@ -924,11 +934,11 @@ joint.dia.CellView = joint.mvc.View.extend({
 
         this.notify('cell:pointerup', evt, x, y);
 
-        if (this._collection) {
+        if (this._graph) {
             // we don't want to trigger event on model as model doesn't
             // need to be member of collection anymore (remove)
-            this._collection.trigger('batch:stop', { batchName: 'pointer' });
-            delete this._collection;
+            this._graph.stopBatch('pointer');
```
kumilingus: For the sake of completeness I would add option `{ cell: this.model }` here.

B3rn475: Done

src/joint.dia.graph.js
```diff
@@ -57,6 +43,8 @@ joint.dia.GraphCells = Backbone.Collection.extend({
 
 joint.dia.Graph = Backbone.Model.extend({
 
+    activeBatches: {},
```
chill117: Can we rename this to `_batches`? I don't think we need to explicitly say that these are "active", and I want it to be clear that this is an internal variable.

B3rn475: done

kumilingus:
Hi Carlo, sorry for the delay. Could you please `rebase` commits into a single commit? So it doesn't pollute the history that much. Thanks.

B3rn475:
Done

kumilingus:
Thank you!

B3rn475:
Are there any info regarding the release of version 0.9.8?

kumilingus:
That should be out before end of the March. I think we are going for `0.10.0` though as there were some new features added.

B3rn475:
Are you going to release some alpha versions via npm? [semver](https://docs.npmjs.com/misc/semver)
It would be great for automatic testing of dependent libraries or applications.

kumilingus:
Could you install JointJS from Master branch directly?
```
npm install git+https://git@github.com/clientIO/joint.git --production
```

Or in `package.json`
```
dependencies: { jointjs: "git+https://git@github.com/clientIO/joint.git" }
```

B3rn475:
Due to #258 I wasn't able to do that, as soon as downloaded the crash of the postinstall script force npm to revert back.
Right now I've created a tarbar with a fix for that and linked to it.
As soon as 0.10.0 comes out I will revert back to the official version.
Thank you for your time.

PR #254, opened 2016-02-19
Performance enhancements 

B3rn475:


src/joint.dia.graph.js
```diff
@@ -950,20 +954,20 @@ joint.dia.Graph = Backbone.Model.extend({
...
-    getBBox: function(cells) {
-        return this.getCellsBBox(cells || this.getElements());
+    getBBox: function(cells, opt) {
+        return this.getCellsBBox(cells || this.getElements(), opt);
     },
 
     // Return the bounding box of all cells in array provided.
     // Links are being ignored.
-    getCellsBBox: function(cells) {
+    getCellsBBox: function(cells, opt) {
```
kumilingus: Oh yeah. We've missed this.

src/joint.dia.paper.js
```diff
@@ -162,10 +162,19 @@ joint.dia.Paper = joint.mvc.View.extend({
...
-        this.listenTo(this.model, 'sort', this.sortViews);
+        this.model.on('add', this.onCellAdded, this);
+        this.model.on('remove', this.removeView, this);
+        this.model.on('reset', this.resetViews, this);
+        this.model.on('sort', function() {
+            if (!this.model.hasActiveBatch('add')) {
+                this.sortViews();
+            }
+        }, this);
+        this.model.on('batch:stop', function(e) {
```
kumilingus: Could you please do some check here that `data` are not `undefined`? I had problem with some compatiblity issues when an application was triggering `batch:stop` without data `{}`.
Also we started calling it `data` instead of `e` or `opt`.

kumilingus:
I can imagine there could be a unit test that would add elements in a batch and `sinon` would spy on `paper.sortViews` to make sure the method was called just once.

B3rn475:
What do you think about the tests I've added?

B3rn475:
Added `sinon` tests

test/jointjs/paper.js
```diff
@@ -22,6 +22,51 @@ module('paper', {
     }
 });
 
+test('paper.addCell() number of sortViews()', function() {
+
+    var callback = sinon.spy();
```
kumilingus: Well instead of triggering a new event (`sorted`) just for the purpose of tests I would rather use `sinon` in the following way.
```
var spy = sinon.spy(this.paper, 'sortViews');
// later
assert.equal(spy.callCount, 1);
```

src/joint.dia.paper.js
B3rn475: I think it was already fixed in #255

kumilingus: Not completely. If an app uses the old approach i.e. `graph.trigger('batch:stop');` (not `graph.stopBatch();`) this code throws an exception.

kumilingus: Note that batch names were added much later than batches themselves.

B3rn475: Sorry I misunderstood the comment. It should be fixed in the last commit.

kumilingus:
@B3rn475  ready to be merged?

B3rn475:
Yes

kumilingus:
Thanks again!

PR #257, opened 2016-02-23
Highlighting of magnet elements now done automatically on paper

chill117:


src/joint.dia.paper.js
```diff
@@ -41,6 +41,70 @@ joint.dia.Paper = joint.mvc.View.extend({
...
+                try {
+
+                    var pathData = magnetEl.convertToPathData();
+
+                } catch (error) {
+
+                    // Failed to get path data from magnet element.
+                    // Draw a rectangle around the entire cell view instead.
+                    magnetBBox = g.rect(magnetBBox);
+                    pathData = V('rect').attr(magnetBBox).convertToPathData();
```
kumilingus: Can we use `V.rectToPath` here and take `rx` & `ry` from `highlightOptions`?

PR #268, opened 2016-03-02
RAP-689: Added "normal" router

chill117:


kumilingus:
Can we add this router as default to the paper?
https://github.com/clientIO/joint/blob/master/src/joint.dia.paper.js#L84

PR #269, opened 2016-03-02
RAP-796: Ensure that mvc.Views' $el has class(es)

chill117:


src/joint.mvc.view.js
```diff
@@ -28,6 +28,20 @@ joint.mvc.View = Backbone.View.extend({
 
         this.setTheme(this.options.theme);
         this.init();
+        this._ensureElClassName();
+    },
+
+    _ensureElClassName: function() {
+
+        var className = typeof this.className === 'function' ? this.className() : this.className;
```
kumilingus: It's better to do `var className = _.result(this, 'className');` here. If the `className` is a function it will be executed in `this` context (that is also what Backbone does).

src/joint.mvc.view.js
```diff
@@ -28,6 +28,20 @@ joint.mvc.View = Backbone.View.extend({
...
+
+        if (className) {
```
kumilingus: The entire block can be rewrite as follows.
```javascript
this.$el.addClass(className); //  jQuery `addClass()` does it all.
```

PR #278, opened 2016-03-19
dia.Paper: grid background takes scale into account

kumilingus:
Not ready for merge. There is an outstanding issue with normalizing dataURL in PhantomJS (possibly related to [this](https://github.com/ariya/phantomjs/issues/10455)).


chill117:
I think it's ok to remove the test samples for scaling; at least until we find a solution to the normalization problem.

kumilingus:
Ready for merge now.

PR #287, opened 2016-03-31
RAP-825: Can no longer create links from magnets when paper.options.i…

chill117:


src/joint.dia.element.js
```diff
@@ -1009,7 +1013,7 @@ joint.dia.ElementView = joint.dia.CellView.extend({
 
             var grid = this.paper.options.gridSize;
             var interactive = _.isFunction(this.options.interactive)
-                ? this.options.interactive(this, 'pointermove')
+                ? this.options.interactive(this)
```
kumilingus: Can we introduce here a new feature e.g. `elementMove` and simply use `fi (this.can('elementMove')) { /* move element */ }`?

src/joint.dia.link.js
```diff
@@ -1446,7 +1436,7 @@ joint.dia.LinkView = joint.dia.CellView.extend({
         // if are simulating pointerdown on a link during a magnet click, skip link interactions
         if (evt.target.getAttribute('magnet') != null) return;
 
-        var interactive = _.isFunction(this.options.interactive) ? this.options.interactive(this, 'pointerdown') : this.options.interactive;
+        var interactive = _.isFunction(this.options.interactive) ? this.options.interactive(this) : this.options.interactive;
```
kumilingus: Can we get rid of this check here and again introduce a new feature? This time e.g `toolsUse` and use it [here](https://github.com/clientIO/joint/blob/master/src/joint.dia.link.js#L1500-L1505) instead?

chill117: I don't quite understand what you want. I don't know enough about the logic within this method to add a `useLinkTools` check as you are suggesting.

kumilingus: First of all I think we should avoid using this construct `var interactive = _.isFunction(this.options.interactive) ? this.options.interactive(this) : this.options.interactive;` on multiple places. As It is already implemented in the `CellView.prototype.can()` method.

When you look at the `pointerdown` method, it stops executing the method rigth at the beginning when `interactive=false`. If there was not such a check the method would carry on executing and terminates with no effect at some other place where `can('feature')` would return `false` (because can(feature) always return `false` if `interactive=false`).

chill117: Okie dokie. So just remove the `var = interactive ..` line? Don't add any new `can()` instances there?

kumilingus: Right. Not there but [here](https://github.com/clientIO/joint/blob/master/src/joint.dia.link.js#L1500-L1505). This is only part that is not inside a feature check. `can('useLinkTools')` should be used here.

chill117: I still don't understand what you want me to change here. Can you do merge this PR then do what you want in a new PR?

kumilingus: Sure :)

PR #290, opened 2016-04-01
dia.LinkView: refactor available magnets marking (use highlighters)

kumilingus:


src/joint.dia.link.js
```diff
@@ -1379,28 +1379,32 @@ joint.dia.LinkView = joint.dia.CellView.extend({
 
     _markAvailableMagnets: function() {
 
-        var elements = this.paper.model.getElements();
-        var validate = this.paper.options.validateConnection;
-
-        _.chain(elements).map(this.paper.findViewByModel, this.paper).each(function(view) {
+        function isMagnetAvailable(view, magnet) {
```
chill117: Defining functions within functions should be avoided. Can this instead be defined on the prototype or within a closure?

PR #312, opened 2016-05-05
disconnect/connect events on link

vtalas:


src/joint.dia.link.js
```diff
@@ -1423,13 +1423,30 @@ joint.dia.LinkView = joint.dia.CellView.extend({
...
+
     startArrowheadMove: function(end, opt) {
+
         opt = _.defaults(opt || {}, { whenNotAllowed: 'revert' });
         // Allow to delegate events from an another view to this linkView in order to trigger arrowhead
         // move without need to click on the actual arrowhead dom element.
         this._action = 'arrowhead-move';
         this._whenNotAllowed = opt.whenNotAllowed;
         this._arrowhead = end;
+        this._initialMagnet = this._getMagnet(end);
```
kumilingus: Does `this[end + 'Magnet'] || this.el` do the same job?

src/joint.dia.link.js
```diff
@@ -1773,6 +1795,18 @@ joint.dia.LinkView = joint.dia.CellView.extend({
                 this._z = null;
             }
 
+            var currentEnd = this.model.prop(arrowhead) || {};
+            var changed = initialEnd.id !== currentEnd.id || initialEnd.port !== currentEnd.port;
+
+            if (changed) {
+                if (initialEnd.id || initialEnd.port) {
+                    this.notify('link:disconnect', evt, arrowhead, this.paper.getModelById(initialEnd.id), this._initialMagnet);
```
kumilingus: This should be `evt, arrowhead, elementView, magnetEl`. Not `evt, arrowhead, elementModel, magnetEl`.

src/joint.dia.link.js
```diff
@@ -1773,6 +1795,18 @@ joint.dia.LinkView = joint.dia.CellView.extend({
                 this._z = null;
             }
 
+            var currentEnd = this.model.prop(arrowhead) || {};
+            var changed = initialEnd.id !== currentEnd.id || initialEnd.port !== currentEnd.port;
```
kumilingus: This is not safe enough. e.g. a change is detected for `{ id: 'a', port: null }` ==> `{ id: 'a' }`.

src/joint.dia.link.js
vtalas: `this.el` is link element in this case, we need source/element

kumilingus: Oh yeah. How about `this[end + 'View'].el`?

vtalas: yes that could works as well :+1:

PR #322, opened 2016-05-11
Make it easy to add ports to any shape

vtalas:
DO NOT MERGE 

thoughts ? comments ? 

example: 
```javascript
var fu = function(index) {

    return { x: index * 10 + 5, y: index * 10 + 0.5 + 5 };
};

var mod = new joint.shapes.basic.Model({
    position: { x: 100, y: 50 },
    size: { width: 100, height: 100 },
    ports: [
        {
            id: 'a',
            position: [0.5, 0.5],
            attrs: {
                rect: { fill: '#ffffff', stroke: '#000000', width: 20, height: 20 }
            },
            markup: '<rect/>'
        },
        {
            id: 'b',
            attrs: { circle: { fill: 'red', r: 10, magnet: true } },
            position: 'left'
        },
        {
            id: 'bx',
            attrs: { circle: { fill: 'blue', r: 5, magnet: true } },
            position: fu //index 0
        },
        {
            id: 'c',
            attrs: { circle: { fill: 'red', r: 10, magnet: true } },
            position: { name: 'left' }
        },
        {
            id: 'd',
            attrs: { circle: { fill: 'red', r: 10, magnet: true } },
            position: 'left'
        },
        {
            id: 'cc',
            attrs: { circle: { fill: 'black', r: 10, magnet: true } },
            position: 'right'
        },
        {
            id: 'xxbx',
            attrs: { circle: { fill: 'blue', r: 5, magnet: true } },
            position: fu //index 1
        },
        {
            id: 'aabx',
            attrs: { circle: { fill: 'blue', r: 5, magnet: true } },
            position: fu //index 2
        },
        {
            id: 'dd',
            attrs: { circle: { fill: 'black', r: 10, magnet: true } },
            position: 'right'
        }
    ],
    attrs: {
        '.rectangle': {
            fill: 'orange',
            stroke: '#000000',
            width: 100,
            height: 60
        },
        text: { text: 'aaa', magnet: 'passive' }
    }
});
graph.addCell(mod);
```
result http://i.imgur.com/PBYOAUN.png


chill117:
I played around with Port layouts demo (demo/port-layouts.html), found some things:
* In the "default settings" example - making a single link from a port to the target element and then removing all the ports results in the link staying but still being connected to both elements. But if I make more than one link, all the links are removed (except for the last one).
* Could you add some tests for any backwards compatibility that we want to keep? I don't remember exactly which APIs we wanted to keep, but I remember Roman saying there were some.
* Once you are polishing up this PR, could you redo the existing demo/ports.html demo to use the new API?

src/ports.js
```diff
@@ -0,0 +1,440 @@
...
+        portMarkup: '<circle class="base-port" r="10" fill="#000000"/>',
+        portLabelMarkup: '<text class="label"/>',
+
+        /**
+         * @private
+         */
+        _initializePorts: function() {
+
+            /** @type {Object<string, {portElement: V, portLabelElement: V}>} */
+            this._portElementsCache = {};
```
chill117: Could this be set via the extend method a few lines above - like this?
```js
    _.extend(joint.dia.ElementView.prototype, {

        portContainerMarkup: '<g class="port"/>',
        portMarkup: '<circle class="base-port" r="10" fill="#000000"/>',
        portLabelMarkup: '<text class="label"/>',

        _portElementsCache: {},
        // ..
```

I like to know all the keys that are used on the object's namespace.

vtalas:
> Here are some of my notes:
> * ~~In the "default settings" example (demo/port-layouts.html) - making a single link from a port to the target element and then removing all the ports results in the link staying but still being connected to both elements. But if I make more than one link, all the links are removed (except for the last one).~~

fixed - when port is removed, link is removed as well
> * Could you add some tests for any backwards compatibility that we want to keep? I don't remember exactly which APIs we wanted to keep, but I remember Roman saying there were some.

not sure about this one either, however the old implentation should not be affected
> * ~~Once you are polishing up this PR, could you redo the existing demo/ports.html demo to use the new API?~~

done

src/ports.js
```diff
@@ -0,0 +1,501 @@
...
+        },
+
+        getPorts: function() {
+
+            return _.cloneDeep(this.prop('ports/items')) || [];
+        },
+
+        getPort: function(id) {
+
+            return _.find(this.getPorts(), function(port) {
```
kumilingus: find first than clone

src/ports.js
```diff
@@ -0,0 +1,501 @@
...
+                _.each(outboundLinks, function(link) {
+
+                    if (removed[link.get('source').port]) link.remove();
+                });
+            }
+        },
+
+        hasPorts: function() {
+
+            return this.getPorts().length > 0;
```
kumilingus: do not clone

src/ports.js
```diff
@@ -0,0 +1,501 @@
...
+        getPort: function(id) {
+
+            return _.find(this.getPorts(), function(port) {
+                return port.id && port.id === id;
+            });
+        },
+
+        getPortIndex: function(id) {
+
+            return _.findIndex(this.getPorts(), function(port) {
```
kumilingus: do not clone

src/ports.js
```diff
@@ -0,0 +1,501 @@
...
+                errorMessages.push('Element: found id duplicities in ports.');
+            }
+
+            return errorMessages;
+        },
+
+        addPorts: function(ports, opt) {
+
+            if (ports.length) {
+                this.prop('ports/items', this.getPorts().concat(ports), opt);
```
kumilingus: do not deep clone, just shallow clone

src/ports.js
```diff
@@ -0,0 +1,501 @@
...
+            }
+
+            return this;
+        },
+
+        removePort: function(port, opt) {
+
+            opt = opt || {};
+            var ports = this.getPorts();
```
kumilingus: do not deep clone

src/ports.js
```diff
@@ -0,0 +1,501 @@
...
+        },
+
+        /**
+         * @private
+         */
+        _renderPorts: function() {
+
+            var ports = _.groupBy(this.model.portData.getPorts(), 'z');
+
+            _.eachRight(_.keys(ports), function(groupName) {
```
kumilingus: `_.eachRight(ports, function(groupPorts, groupName) { /* .. */ }, this)`

src/ports.js
```diff
@@ -0,0 +1,501 @@
...
+         * @param {Array<Port>}ports
+         * @param {number} z
+         * @private
+         */
+        _appendPorts: function(ports, z) {
+
+            var containerElement = this.rotatableNode || this.vel;
+            var childNodes = containerElement.node.childNodes;
+
+            var getPortElement = function(port) {
```
kumilingus: Can we defined this method on prototype?

plugins/layout/ports/joint.layout.port.js
```diff
@@ -0,0 +1,180 @@
...
+        }, g.line(p1, p2));
+    }
+
+    function ellipseLayout(ports, elBBox, startAngle, stepFn) {
+
+        var center = elBBox.center();
+        var ratio = elBBox.width / elBBox.height;
+        var p1 = g.point(elBBox.width / 2, 0);
+
+        var ellipse = g.Ellipse.fromRect(elBBox);
```
kumilingus: `g.Ellipse` does not exist, right? only `g.ellipse`?

src/geometry.js
```diff
@@ -209,6 +209,24 @@ var g = (function() {
             return Ellipse(this);
         },
 
+        tangentTheta: function(p) {
```
kumilingus: Can we please add a test for this?

src/vectorizer.js
```diff
@@ -117,18 +117,24 @@ V = Vectorizer = (function() {
 
     /**
      * @param {SVGMatrix} matrix
-     * @returns {V, SVGMatrix} Setter / Getter
+     * @param {Object=} opt
+     * @returns {Vectorizer|SVGMatrix} Setter / Getter
      */
-    V.prototype.transform = function(matrix) {
+    V.prototype.transform = function(matrix, opt) {
```
kumilingus: Can we please add a test for the `absolute` option?

plugins/layout/ports/joint.layout.port.js
kumilingus: Please ignore this. The `g.Ellipse` alias for `g.ellipse` has been added.

PR #325, opened 2016-05-15
Geometry: add leftMiddle(), rightMiddle(), topMiddle(), bottomMiddle(…

kumilingus:
…) rect methods, add rect.fromEllipse() ellipse.fromRect(), add point, rect and ellipse defaults

### DIFF
--- a/src/geometry.js
+++ b/src/geometry.js
@@ -48,9 +48,6 @@ var g = (function() {
         } else if (Object(x) === x) {
             y = x.y;
             x = x.x;
-        } else {
-            x = x;
-            y = y;
         }
         this.x = x === undefined ? 0 : x;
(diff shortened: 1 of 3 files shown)


src/geometry.js
```diff
@@ -41,18 +41,19 @@ var g = (function() {
...
         } else if (Object(x) === x) {
-            this.x = x.x;
-            this.y = x.y;
+            y = x.y;
+            x = x.x;
         } else {
-            this.x = x;
-            this.y = y;
+            x = x;
+            y = y;
```
chill117: I don't think these lines are needed anymore.

kumilingus: Good point :)

PR #328, opened 2016-05-21
Added style field  to package.json

mike-goodwin:
To support requiring css like with [npm-css](https://www.npmjs.com/package/npm-css) to make it easier for people to consume the package when they are using npm for all the things instead of gulp/bower/grunt. I chose the minified css because i figured that's what most people would use in production.


chill117:
Looks fine to me.

PR #341, opened 2016-06-13
[SUGGESTION] removed usage of fs module to make it work in the browser

quentin-sommer:
I'm not sure if it would cause any issues


chill117:
Thank you for the PR, but I am already in the process of fixing the issue.

PR #364, opened 2016-07-28
Create `portProp` method 

vtalas:
- portProp allows to set port properties by port id 
- added `data-group` attribute to the port DOM element 

ready @kumilingus

### DIFF
--- a/src/ports.js
+++ b/src/ports.js
@@ -495,8 +495,10 @@
                 throw new Error('ElementView: Invalid port markup - multiple roots.');
             }
 
-            portContentElement.attr('port', port.id);
-            portContentElement.attr('data-group', port.group);
+            portContentElement.attr({
+                'port': port.id,
+                'port-group': port.group
+            });
 
             var portElement = V(this.portContainerMarkup)
                 .append(portContentElement)
--- a/src/ports.js
+++ b/src/ports.js
@@ -239,6 +239,35 @@
             return this;
         },
 
+        /**
+         * @param {string} portId
+         * @param {string} path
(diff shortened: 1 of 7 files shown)


src/ports.js
```diff
@@ -239,6 +239,35 @@
...
+         */
+        portProp: function(portId, path, value, opt) {
+
+            var index = this.getPortIndex(portId);
+
+            if (index === -1) {
+                throw new Error('Element: unable to find port with id ' + portId);
+            }
+
+            if (!_.startsWith(path, '/')) {
```
kumilingus: Note that if `path` starts with `/` it points to a different property in an object.

`'a/b'    { a: { b: 'value' }`

`'/a/b'   { '': { a: { b: 'value' }}`

vtalas: in this case at least one `/` has to be there, as the `path` is just a part of the 'result' path, 
` var portPath = ['ports/items/', index, path].join('');`

kumilingus: Sure. The `/` has to be there in any case. Even if the `path` already starts with `/`.

    var portPath = ['ports/items/', index, '/', path].join('');

src/ports.js
```diff
@@ -239,6 +239,35 @@
...
+                path = '/' + path;
+            }
+
+            var portPath = ['ports/items/', index, path].join('');
+
+            if (_.isUndefined(value)) {
+                return this.prop(_.trimRight(portPath, '/'));
```
kumilingus: The same applies here. You can not trim last `/`, as you would change the path meaning.

PR #378, opened 2016-08-09
core: get rid of the mixin, supplement methods

vtalas:
- supplement -> _.defaults 
- deepSupplement -> _defaultsDeep
- mixin -> _.extend
- deepMixin -> _.merge


chill117:
Looks good to me: all tests still pass, and all internal usage of those four methods has been removed.

Are you we waiting with this for some reason?

PR #423, opened 2016-09-22
responsive navigation menu

pskala:


pskala:
removed to jointjs_site

PR #434, opened 2016-10-05
exclude everything but dist folder in bower file

### LINKED ISSUE #429: Ignore docs, demo, source and test folders in bower.json
Actually everything except dist and maybe a couple of other folders should be ignored

alirezamirian:
closes #429


kumilingus:
Thanks!

PR #473, opened 2016-11-22
Fix jumpover on cells reset

HiaslTiasl:
When cells are reset, only the local variable got replaced and the old list of views was used in `updateJumpOver()`.

The symptoms were errors thrown on every `stop:batch` event resulting from missing `ownerSVGElement`s.


kumilingus:
Good catch! Thank You.

PR #476, opened 2016-11-26
ports: getPortsPositions

vtalas:
- `getPortsPositions(groupName)` retrieves retrieve port positions even if element is not rendered yet 
- fix the `portProp(id)` - it should get all port properties
- refactor each -> transform where possible

ready for review

### DIFF
--- a/src/ports.js
+++ b/src/ports.js
@@ -39,14 +39,14 @@
             }
         },
 
-        getLayoutPosition: function (id) {
+        getLayoutPosition: function(id) {
 
-            var port= this.getPort(id);
-            if (port) {
-                return port.layoutPosition;
+            var port = this.getPort(id);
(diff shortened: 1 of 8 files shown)


src/ports.js
```diff
@@ -31,6 +31,24 @@
             this.ports.push(port);
         },
 
+        storeLayoutPosition: function (id, position) {
```
DavidDurman: shouldn't this be called `setLayoutPosition()`? I guess we have "getters" and "setters", not "getters" and "storers".

vtalas: `set` seems to me that it could affect the port position, in this case value is just stored for later use, port position is not affected.

vtalas: i'v changed it to `set` anyway ...

PR #482, opened 2016-11-30
maybe a bug in joint.layout.DirectedGraph.js?

lialosiu:
Why ignore `opt.setLinkVertices` and call `opt.setVertices` ?

The [document](http://resources.jointjs.com/docs/jointjs/v1.0/joint.html#layout.DirectedGraph) say we should use `setLinkVertices`... 　Σ(ﾟДﾟ)

### DIFF
--- a/plugins/layout/DirectedGraph/joint.layout.DirectedGraph.js
+++ b/plugins/layout/DirectedGraph/joint.layout.DirectedGraph.js
@@ -108,14 +108,12 @@ joint.layout.DirectedGraph = {
                 var points = glEdge.points || [];
 
                 if (opt.setLinkVertices) {
-                    if (opt.setVertices) {
-                        opt.setVertices(link, points);
-                    } else {
-                        // Remove the first and last point from points array.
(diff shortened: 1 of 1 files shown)


kumilingus:
Hi,

`setLinkVertices` option  is a `boolean` - telling whether you want the layout to set the vertices on the links or not.

`setVertices` option is a method to set actual vertices on links. For instance you can use transitions here to change the positions of vertices in animated manner.

With your solution one would have to pass a "no operation" method as `setVertices` when setting vertices is not desired  i.e.  `setVertices: function() { /* noop */ }`

lialosiu:
Yeah, I see...
I just try to use "setLinkVertices" just like document usage, and found it not work, so I doubt of it is a bug or not.

kumilingus:
I see! The docs are wrong. We should update them instead. Or change the behaviour little.

`setLinkVertices: false` - do not set vertices
`setLinkVertices: true` - set vertices using default approach
`setLinkVertices: function() {}` - use the function

PR #501, opened 2017-01-14
dia.ElementView: stop temporary removing the 'translate' attribute fo…

kumilingus:
…r bbox calculation (subtract the translate from the bbox position)


kumilingus:
Fixed in #502.

PR #509, opened 2017-02-10
dia.Cell: prop(), removeProp() should accept path defined as an array

vtalas:
ready for review

### DIFF
--- a/src/joint.dia.cell.js
+++ b/src/joint.dia.cell.js
@@ -420,13 +420,14 @@ joint.dia.Cell = Backbone.Model.extend({
                 }
 
                 var property = pathArray[0];
+                var pathArrayLength = pathArray.length;
 
                 opt = opt || {};
                 opt.propertyPath = path;
                 opt.propertyValue = value;
                 opt.propertyPathArray = pathArray;
 
-                if (pathArray.length === 1) {
+                if (pathArrayLength === 1) {
                     // Property is not nested. We can simply use `set()`.
                     return this.set(property, value, opt);
                 }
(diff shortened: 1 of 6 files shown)


src/joint.dia.cell.js
```diff
@@ -504,7 +518,25 @@ joint.dia.Cell = Backbone.Model.extend({
     removeAttr: function(path, opt) {
 
         if (_.isArray(path)) {
-            _.each(path, function(p) { this.removeAttr(p, opt); }, this);
+
+            var clone = path.slice();
```
kumilingus: Do you have any special reason why you do not use `['attrs'].concat(path)` here? It's not slower (https://jsperf.com/array-unshift-vs-prepend/5) and easier to read IMO.

src/joint.dia.cell.js
```diff
@@ -504,7 +518,25 @@ joint.dia.Cell = Backbone.Model.extend({
...
+            clone.unshift('attrs');
+            return this.removeProp(clone);
+        }
+
+        return this.removeProp('attrs/' + path, opt);
+    },
+
+    removeAttrs: function(path, opt) {
```
kumilingus: The method supposed to remove multiple attributes i.e. `removeAttrs(paths, opt)`.  If `paths` is a string do not remove single attribute. That's why `removeAttr()` is there.

src/joint.dia.cell.js
vtalas: good point, totally agreed, I'll change it in method `attrs` as well

src/joint.dia.cell.js
vtalas: ok, it will throw an error path is not an array, also `path` should be `attrs` instead

PR #518, opened 2017-02-20
dia.Element.port: add:ports, remove:ports events proposal 

vtalas:
with only `change:ports` event it's hard to detect when ports are added or removed to the element.  

`add:port/remove:port` events provide the `cell` and array of added/removed elements
```
element.on('add:ports', function(cell, addedPorts) {
    console.log("add port(s)", arguments);
});

element.on('remove:ports', function(cell, removecPorts) {
    console.log("remove port(s)", arguments);
});
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,6 @@
+??-??-???? (v1.1.0)
+ * ports events `ports:add` and `ports:remove` triggered when port is added to element/removed from element
+
 18-11-2016 (v1.0.3)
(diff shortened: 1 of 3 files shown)


kumilingus:
`element.portProp('id-old', 'id', 'id-new')` will not trigger `port:removed` and `port:added`. Is that intended behaviour? What about `element.prop('ports/items', [{ id: 'x' }, { id: 'y' }], { rewrite: true })` when there were 2 ports already defined.

PR #521, opened 2017-02-20
Creates initial typescript definitions right in the JointJS project.

mmacfadden:
This was merged from what was in Definitely Typed.  However, what we there was not modular and was missing several things like the mvc package.  These need to be continually improved but I think is a great start.  I am happy to continually evolve this.


mmacfadden:
Ok, I think this roughly works. Everything that was in the definitely typed version is included here plus a lot more that was missing.  There is still a bunch of the API that is not typed, and I suspect that we will need to go through the API for consistency.  But I think this is a good start.

PR #523, opened 2017-02-22
dia.Paper: SVG grid

vtalas:
ready for review !

 - added possibility to create custom grid shapes  - defined as the svg markup. 
 - removed drawing into canvas - serialize svg markup then use the base64 value for the `.joint-paper-grid` element

a bit faster, more precise - with canvas drawn grid didn't match the paper grid for some combinations of scale and gridsize


src/joint.dia.paper.js
```diff
@@ -1374,27 +1374,84 @@ joint.dia.Paper = joint.mvc.View.extend({
...
+
+        var svgGridElements = this._svgGridData;
 
-        opt = _.defaults({}, opt, this.options.drawGrid);
+        if (!svgGridElements) {
+
+            svgGridElements = {
+                xmlSerializer: new XMLSerializer(),
+                pattern: V('<pattern id="grid-pattern" patternUnits="userSpaceOnUse"></pattern>'),
+                gridShape: opt.gridPattern ? V(opt.gridPattern) : V('<rect />').attr('fill', opt.color)
```
kumilingus: Can we move "dot" logic to default options as  `gridPattern` and `updateGridPattern`?

demo/paper/src/paper.js
```diff
@@ -4,6 +4,26 @@ var paper = new joint.dia.Paper({
...
+        /**
+         * define custom grid pattern shape
+         */
+        // gridPattern: '<rect width="100%" height="100%" stroke="red" stroke-width="1" fill="transparent"/>',
+
+        /**
+         * define custom shape with updating attributes
+         */
+        gridPattern: '<path stroke="green" stroke-width="1"/>',
+        updateGridPattern: function(el, opt) {
```
kumilingus: Please always expose `SVGElement|HTMLElement` not Vectorizer element nor jQuery element.

kumilingus:
image
Can we create grid like that? e.g. every n-th line is thicker.

src/joint.dia.paper.js
vtalas: Yes, it will be better. I was trying to minimize  `upadatePattern` call, However, there is `upadtePattern` even for the default  'dot'.

demo/paper/src/paper.js
vtalas: ok

PR #524, opened 2017-02-23
Update dist: add typescript definition file

kumilingus:


mmacfadden:
@kumilingus This would work, but one other small change would be required as well.  The package.json would need the following entry.

"types": "./dist/joint.d.ts"

mmacfadden:
@kumilingus Oh wait.  I think we already added that.  I forgot that the package.json sits in the root of the project, not in the dist folder.

PR #525, opened 2017-02-23
Improves typescript definitions

mmacfadden:
The original stab at this exported JointJS as a module.  However, jointjs is primarily exported into the global namespace as the "joint" variable. The changes here make it easier to augment the Typescript definitions in the jointjs project with additional classes, etc in rappid, and have them all fall under the "joint" global variable.  If we also need to export as a module later, we can do that to.

I noticed that the original DefinitelyTyped project also did it the way I am now committing, so in that way it will not break any one else who was using the DefinitelyTyped version before.  The commit also updates the `dist/joint.d.ts` file.


mmacfadden:
Actually, I am still experimenting with this.  It might be worth a conversation before merging this.

doender:
I don't whether this is useful for you, but I had the following errors with an existing project (working in 1.0.2) but trying to use the typescript definitions:

```
Property 'TextBlock' does not exist on type 'typeof basic'.

Property 'extend' does not exist on type 'typeof Link'.

Argument of type '{ setLinkVertices: boolean; }' is not assignable to parameter of type 'LayoutOptions'.
  Types of property 'setLinkVertices' are incompatible.

Type 'boolean' is not assignable to type '(link: Link, vertices: Position[]) => void'.

Argument of type '{ el: any; model: any; width: string; height: number; gridSize: number; drawGrid: { color: string...' is not assignable to parameter of type 'PaperOptions'.
  Object literal may only specify known properties, and 'drawGrid' does not exist in type 'PaperOptions'.)

Property 'model' does not exist on type 'CellView'.
```

mmacfadden:
Thanks, I will update the typescript definitions.

vtalas:
Hi Micheal, 
First of all -  thank you for a great job of creating these definitions.

 I've managed to get running a basic demo, however, I had to put back the `index.d.ts` definition. Without that, I wasn't able to compile the demo example in `demo/ts-demo`. Could you please provide an example or a snippet how do use it without the `index.d.ts`?   

my demo solution is in the https://github.com/vtalas/joint/tree/ts

mmacfadden:
Hello. I just saw this comment. Did you figure it out or do you still need some help?

vtalas:
I refactored it slightly, at the moment it seems it's working. 

I have a usage of the joint.d.ts, something like 

```
import * as joint from "../../dist/joint";
import * as $ from "jquery";

const graph = new joint.dia.Graph;
const paper = new joint.dia.Paper({
    el: $('#paper'),
    width: 650,
    height: 400,
    gridSize: 20,
    model: graph,
    markAvailable: true,
    linkConnectionPoint: joint.util.shapePerimeterConnectionPoint
});

const a = new joint.shapes.basic.Rect({
    position: {x: 50, y: 50},
    size: {width: 100, height: 40},
    attrs: {text: {text: 'basic.Rect'}}
});
graph.addCell(a);

````

In your PR you removed the `index.d.ts` file. I'm curious how do you use the `dist/joint.d.ts` as I wasn't able to get it working without the `export` statement.

PR #537, opened 2017-03-10
Replaced typings with DefinitelyTyped typings

FBeuselinck:
I modified some of the original typings files with typings from "https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/jointjs". In the project i'm using this typescript now recognizes a lot more definitions such as the ones from the extra plugins (logic, chess, devs, ...).


kumilingus:
Hi, has `package.json` been commited here by a mistake?

FBeuselinck:
Yes, i think that's a mistake. I can't recall that I changed the package.json in the root folder. I did however add a package.json file to the "types" folder.

PR #549, opened 2017-03-21
Vectorizer: improve convertRectToPathData() for rounded rectangles

kumilingus:


kumilingus:
Good point! I've moved all the logic to `V.rectToPath`.

PR #570, opened 2017-04-11
Flesh out joint.dia types. Clarify options on methods where I could understand them

ChrisMoran:
This adds some missing method definitions on the Cell, CellView, Element, ElementView, Link, LinkView, Graph, and Paper classes.

I've tried to clarify some of the options types that are passed into most method. I've switched uses of any over to object, since many of those uses were actually expecting an object type, not a number or string. Tested by building with grunt and compiling with tsc == 2.2.0

### DIFF
--- a/types/joint.d.ts
+++ b/types/joint.d.ts
@@ -35,14 +35,18 @@ export namespace dia {
         deep?: boolean;
     }
 
+    interface EdgeMap {
+	[key: string]: boolean;
+    }
+
     class Graph extends Backbone.Model {
-        constructor(attributes?: any, options?: { cellNamespace: any });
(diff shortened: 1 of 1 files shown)


chill117:
Thank you for the efforts you've put into this. I don't have much experience with typescript. How would you recommend adding tests for the type definitions in this project?

ChrisMoran:
@chill117 No problem! 

I've not added tests before for type definitions like this, but I would think having typescript code that calls the type definitions would be a good start. If there's something missing or incorrect, then the typescript compiler will complain about it and not compile or, at the very least, emit warnings.

vtalas:
would be great to (re)write unit tests in TS

PR #571, opened 2017-04-12
fix typescript errors

tiaguinho:
fix typescript 2.2.1 errors compilation


tiaguinho:
Ok, there is something wrong in generated file.
I check the files under types/ and the params are typed, but in the generated file aren't.

I think just compile the files again will fix the bug.

Thanks for your attention.

PR #573, opened 2017-04-12
TS def: typescript definition cleanup 

vtalas:
remove private methods from definition files 
- [x]  CellViewGeneric
- [x] CellView
- [x] ElementView
- [x] dia.Link
- [x] dia.Element 
- [x] dia.Paper
- [x] dia.Graph

### DIFF
--- a/demo/ts-demo/tsconfig.json
+++ b/demo/ts-demo/tsconfig.json
@@ -1,7 +1,7 @@
 {
 	"compilerOptions": {
 		"module": "es6",
-		"noImplicitAny" : false
+		"noImplicitAny" : true
 	},
 	"compileOnSave": false
 }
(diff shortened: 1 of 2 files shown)


ChrisMoran:
Don't you need some of these methods defined if you wanted to extend this class in typescript?

ChrisMoran:
To clarify what I mean, say you wanted to create a custom Element and ElementView like:

```
class CustomElement extends ElementView {
  markup = '<rect class="regular"/><rect class="special"/>'
  ...
}

class CustomElementView extends ElementView {
  mouseover: (evt: Event) => {
    if (evt.target.classList.contains('special')) {
      this.notify('cell:mouseoverspecial', evt);
    } else {
      super.mouseover(evt);
    }
  }
}
```

Now the Typescript compiler wouldn't see the mouseover method on ElementView, and the above wouldn't compile.

vtalas:
oh yes, good point! it totally makes sense leave these there, check out my latest updates, `protected` could be the right way, what do you think?

ChrisMoran:
Protected makes much more sense, thanks for updating! This change 👍

PR #593, opened 2017-05-13
add LinkView supp for findView and findViewByModel

mahald:
added missing Link/LinkView support for paper.findView and
paper.findViewByModel

### DIFF
--- a/types/joint.d.ts
+++ b/types/joint.d.ts
@@ -501,9 +501,9 @@ export namespace dia {
         drawGrid(options?: {width?: number, height?: number, scaleFactor?: number,
                             update: any, ox?: number, oy?: number}): this;
 
-        findView<T extends CellView | LinkView>(element: string | JQuery | SVGElement): T;
+        findView<T extends ElementView | LinkView>(element: string | JQuery | SVGElement): T;
 
-        findViewByModel<T extends CellView | LinkView>(model: Cell | string | Link) : T;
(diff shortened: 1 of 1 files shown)


kumilingus:
How about the `Element`/`ElementView`? Both `Element` and `Link` inherit from `Cell`. Is this necessary? Thanks.

mahald:
ok it's wrong anyways, it would have to be: "ElementView | LinkView".

imho if using typescript it should return precise types and not kinda base class. 
for sure u can cast it but then u as good could use any as type.

currently i have to do it as follows:
`
this.graph.on('change:position', function(cell) {
    let links: joint.dia.Link[] = this.graph.getLinks()
	for (let i = 0; i < links.length; i++) {
        let x = paper.findViewByModel(links[i]) as joint.dia.LinkView
        x.update(null, null);
	};
}.bind(this));
`

i would prefer:
`paper.findViewByModel<joint.dia.LinkView>(links[i]).update(null, null); `
instead

the big advantage in the second case is, the editor is going to tell me what types are possible (intelli Sense). In the first case i got to consult the manual or guess it.

just my 2 cents

kumilingus:
It makes totally sense to me. If you can explicitly add both `ElementView` and `LinkView` to the definition I am happy to merge this. Thank You.

mahald:
Ok updated.

kumilingus:
Thanks!

PR #601, opened 2017-05-25
Fix toolMarkup typing typo

pruggia:


kumilingus:
Thanks!

pruggia:
Thanks for the approval. Quick question, the one in the dist directory still has the typo, will that get automatically fixed when you do a new release or I should have fixed that one too?
Also, do you have a fixed release cycle or you do adhoc releases when there are enough features/fixes that makes it worth it?

kumilingus:
Right, it's more an ad-hoc release. The `dist` directory will be updated when there is a new release by running `grunt dist`.

PR #609, opened 2017-06-14
Changed Deprecated xlink:href to svg2 href.

GooseNinja:
Not sure if this is something you want for joint.js, but 
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href


kumilingus:
Hi, unfortunately `href` attribute on SVG image element is not supported by Safari yet (see Puzzle demo for instance).

GooseNinja:
I didn't know that. My bad :)

kumilingus:
More of a Safari bad :) Thank you anyway.

PR #614, opened 2017-06-23
Vectorizer: refactor findIntersection() - WIP

kumilingus:


kumilingus:
See https://github.com/clientIO/joint/pull/838.

PR #615, opened 2017-07-03
dia.Cell: stop merging arrays attributes inside of the constructor

kumilingus:
```javascript
var MyCell = joint.dia.Cell.extend({ attribute: ['value'] });
var myCell = new MyCell({ attribute: [] });
// Previously
console.log(myCell.get('attribute')) // output `['value']`
// After this change
console.log(myCell.get('attribute')) // output `[]`
```
* [ ] make it optional to easily maintain backwards compatibility
* [ ] define the customizer function on prototype/constructor to avoid function creation when constructing a cell


kumilingus:
Addressed in https://github.com/clientIO/joint/pull/2455.

PR #700, opened 2017-10-10
TS def: fix mvc.View 

vtalas:
tabs -> spaces in package.jsons
add package-lock in ts-demo


types/joint.d.ts
```diff
@@ -1243,3 +1340,33 @@ export namespace layout {
         export function layout(graph: dia.Graph | dia.Cell[], options?: LayoutOptions): dia.BBox;
     }
 }
+
+export namespace mvc {
+
+    interface View extends Backbone.View<Backbone.Model>, Backbone.Events {
```
mmacfadden: At a minimum, I believe this line needs to be:

```TypeScript
interface View<T extends Backbone.Model> extends Backbone.View<T>, Backbone.Events {
```
We still need mvc.View to be generic, because the Paper class needs to extend Backbone.View with the Graph object for its model, not a generic Backbone.Model.

The other concern is that this is not an interface, it is a class.  Why are we defining it as an interface?

```TypeScript
class View<T extends Backbone.Model> extends Backbone.View<T>, Backbone.Events {
```

types/joint.d.ts
```diff
@@ -454,299 +452,310 @@ export namespace dia {
...
 
-        setInteractivity(value: any): void;
+        protected pointermove(evt: Event, x: number, y: number): void;
 
-        setOrigin(x: number, y: number): this;
+        protected pointerup(evt: Event, x: number, y: number): void;
+    }
 
-        snapToGrid(x: g.Point | number, y?: number): g.Point;
+    interface Paper extends mvc.View {
```
mmacfadden: Paper is not an interface, it is a class.  Also if we make mvc.View generic (see below) this should be:

```TypeScript
class Paper extends mvc.View<Graph> {
```

types/joint.d.ts
```diff
@@ -454,299 +452,310 @@ export namespace dia {
...
 
-    class Paper extends Backbone.View<Graph> {
-        constructor(options?: PaperOptions);
+    var Paper: PaperStatic;
 
-        options: PaperOptions;
-        svg: SVGElement;
-        viewport: SVGGElement;
-        defs: SVGDefsElement;
+    interface PaperStatic {
```
mmacfadden: If we revert this back to a class, then we don't need this, and should put the constructor back.

vtalas:
There is another product - RappidJs based on the JointJs library. In RappidJS, there are some extensions made on the `dia.Paper`, e.g. : 
```
joint.dia.Paper.prototype.print = function(opt) {
...
```

the motivation for all of this "interfacing" was to make the `dia.Paper` extendible in another d.ts file. TS does not support partial classes (https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods)

however, interfaces can be defined partially, this definition is valid:
```
interface Paper {
  a: void; 
}
interface Paper {
  print: void
}
```

vtalas:
#711

mmacfadden:
@vtalas I see what you are trying to do now. I will have to think about it a little bit. The reason why interfaces become problematic is because in typescript you cannot do "instanceof" on an interface. So if we define them that way the following code would not compile:

```TypeScript
if (foo instanceof joint.dia.Paper) {
}
```

mmacfadden:
@vtalas I think you want to use class augmentation via interfaces.  So for example this is perfectly valid:

```TypeScript
export class Paper {
  public bar(): void {
  }
}

export interface Paper {
  baz(): void;
}

const paper: Paper = new Paper();
paper.bar();
paper.baz();
```

Paper is defined as a **class** but I then augment the paper definition with an interface, and the declarations are merged, but the end result is a class. This could happen in different files, so perhaps in joint, you define the class, and in rappid you augment with an interface?

https://www.typescriptlang.org/docs/handbook/declaration-merging.html

vtalas:
Works as a charm! thank you for that.

Issue #677, opened 2017-09-21
cell.id must be string???


munepom:
I noticed... When cell.id is number,  the following exception occurs...

test code
```
<!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8">
    <title>Issue</title>
    <link rel="stylesheet" href="assets/rappid/dist/rappid.min.css" />
</head>
<body>
    <div id="Paper" class="joint-theme-default joint-paper"></div>

    <!-- Dependencies: -->
    <script src="assets/rappid/node_modules/jquery/dist/jquery.js"></script>
    <script src="assets/rappid/node_modules/lodash/index.js"></script>
    <script src="assets/rappid/node_modules/backbone/backbone.js"></script>
    <script src="assets/graphlib/graphlib.min.js"></script>
    <script src="assets/dagre/dagre.js"></script>
    <script src="assets/rappid/dist/rappid.min.js"></script>

    <!-- Apps: -->
    <script>
var graph = new joint.dia.Graph()
var paper = new joint.dia.Paper({
    width: 200,
    height: 200,
    gridSize: 1,
    model: graph,
    snapLinks: false,
    linkPinning: false,
    linkConnectionPoint: joint.util.shapePerimeterConnectionPoint,
    interactive: function(cellView, method) {
        return false
    }
})
var paperScroller = new joint.ui.PaperScroller({
    paper: paper,
    cursor: 'grab',
    autoResizePaper: true,
    baseWidth: 1,
    baseHeight: 1,
    contentOptions: {
        padding: 100,
        allowNewOrigin: 'any'
    }
})

paperScroller.render().$el.appendTo('#Paper')

var rect1 = new joint.shapes.basic.Rect({
  position: { x: 50, y: 50 },
  size: { width: 50, height: 50 },
}).set('id', 1).addTo(graph)

var rect2 = new joint.shapes.basic.Rect({
  position: { x: 150, y: 150 },
  size: { width: 50, height: 50 },
}).set('id', 2).addTo(graph)

var link = new joint.dia.Link({
  source: { id: rect1.id },
  target: { id: rect2.id }
}).addTo(graph) // error occurs

    </script>
</body>
</html>
```

```
rappid.min.js:56 Uncaught TypeError: Cannot read property 'el' of undefined
    at child.onEndModelChange (rappid.min.js:56)
    at child.wrapper (index.js:3592)
    at child.b [as watchSource] (rappid.min.js:56)
    at child.render (rappid.min.js:55)
[...]

PR #704, opened 2017-10-12
fix paper.findViewByModel()

munepom:
# Issue
https://github.com/clientIO/joint/issues/677

# Summary
- When typeof cell is number, then error occurs... So I changed `paper.findViewByModel()`.
- I also make `paper.findViewByModel()` not throwing an exception when called with a falsy value.
~~- (We can use boolean values as a side effect... Shoud we check it?)~~

Please review this p-r :)

### DIFF
--- a/src/joint.dia.paper.js
+++ b/src/joint.dia.paper.js
@@ -855,7 +855,7 @@ joint.dia.Paper = joint.mvc.View.extend({
     // Find a view for a model `cell`. `cell` can also be a string or number representing a model `id`.
     findViewByModel: function(cell) {
 
-        var id = joint.util.isObject(cell) ? cell.id : cell;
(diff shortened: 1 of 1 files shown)


kumilingus:
> I also make paper.findViewByModel() not throwing an exception when called with a falsy value.

To fulfil the above we might change it to the following:
```javascript
var id = (joint.util.isString(cell) || joint.util.isNumber(cell)) ? cell : (cell && cell.id);
```

munepom:
oh! null or undefined check is needed 💦 
I'll fix it.

munepom:
Thank you for fast review!

# Changed
- check null or undefined to avoid exception

kumilingus:
Thank you!

PR #736, opened 2017-10-20
dia.Paper: update TS definitions

kumilingus:


types/joint.d.ts
```diff
@@ -532,20 +539,31 @@ export namespace dia {
...
+        image?: string;
+        quality?: number;
+        position?: Point | string;
+        size?: Size | string;
+        repeat?: string;
+        opacity?: number;
+        waterMarkAngle?: number;
+    }
+
+    interface PaperOptions extends mvc.ViewOptions<Graph> {
```
vtalas: wouldn't be better to have something like:
```
namespace Paper {
    export interface Options {
        // ...
    }
    export interface GridOptions {
        // ...
    }
    export interface BackgroundOptions {
        // ...
    }
}
```
?

kumilingus: Definitely. I'll make changes.

PR #767, opened 2017-11-20
Avoid to change z values during toFront or toBack

B3rn475:
Two consecutive invocations of toFront or toBack produce a change in the
cells even though they already have the correct z values.

If the z values already satisfy the required order for toFront or toBack
they are not changed.

__Context:__
In an application I'm developing every time a user clicks on a cell i bring it to front, if the user starts clicking on the same cell over and over i get overwhelmed by `change:z` events.

### DIFF
--- a/src/joint.dia.cell.js
+++ b/src/joint.dia.cell.js
@@ -178,18 +178,34 @@ joint.dia.Cell = Backbone.Model.extend({
 
             opt = opt || {};
 
-            var z = (this.graph.getLastCell().get('z') || 0) + 1;
+            var z = this.graph.maxZIndex();
 
-            this.startBatch('to-front').set('z', z, opt);
+            var cells;
 
             if (opt.deep) {
-
-                var cells = this.getEmbeddedCells({ deep: true, breadthFirst: true });
-                cells.forEach(function(cell) { cell.set('z', ++z, opt); });
-
+                cells = this.getEmbeddedCells({ deep: true, breadthFirst: true });
+                cells.unshift(this);
+            } else {
+                cells = [this];
             }
 
-            this.stopBatch('to-front');
+            z = z - cells.length + 1;
+
+            var shouldUpdate = cells.some(function(cell, index) {
+                return cell.get('z') !== z + index;
+            });
+
+            if (shouldUpdate) {
+                this.startBatch('to-front');
+
+                z = z + cells.length;
+
+                cells.forEach(function(cell, index) {
+                    cell.set('z', z + index, opt);
+                });
+
+                this.stopBatch('to-front');
+            }
         }
 
         return this;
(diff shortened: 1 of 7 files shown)


kumilingus:
Hi Carlo, thanks for the PR.  We're in the middle of a transition to remove the Lodash dependency. We do not call `lodash` methods directly since `v2.0` . Please use `joint.util.*()` instead (https://github.com/clientIO/joint/blob/master/src/core.js#L1423-L1452) or even better use ES5 only. i.e.
```javascript
// head
array.slice(0, count)
// tail
array.slice(-count)
```

B3rn475:
I refactored as requested

B3rn475:
@kumilingus any update?

If you have any other comment I'll be happy to address them.

kumilingus:
Hi Carlo, sorry for the delay.

Is it necessary to loop over all cells in the graph to get the highest z-index?
```javascript
var z = this.graph.getCells().reduce(function (maxZ, cell) {
  return Math.max(maxZ, cell.get('z'));
}, -Infinity);
```

All the cells in the graph are sorted by `z` already.
```javascript
var z = this.graph.getLastCell().get('z') || 0;
```

Thank you.

src/joint.dia.cell.js
```diff
@@ -201,17 +219,34 @@ joint.dia.Cell = Backbone.Model.extend({
...
-                var cells = this.getEmbeddedCells({ deep: true, breadthFirst: true });
-                cells.reverse().forEach(function(cell) { cell.set('z', z--, opt); });
+                cells = this.getEmbeddedCells({ deep: true, breadthFirst: true });
+                cells.unshift(this);
+            } else {
+                cells = [this];
             }
 
-            this.set('z', z, opt).stopBatch('to-back');
+            var shouldUpdate = cells.reduce(function(shouldUpdate, cell, index) {
```
kumilingus: Using `cells.some()`might be more efficient here. (It should be supported from IE9+).

B3rn475:
I see now that there is a method `maxZIndex` on the `Graph` object to do it already, so I should probably use it.
https://github.com/clientIO/joint/blob/b1c5763ae5d1887644a9a0ac1c4845a7edfe2b5a/src/joint.dia.graph.js#L296

The problem is related to batches.
The cells are sorted at the end of the batch so if we query the "last cell" during an already running batch it may not be the correct value we were expecting.
https://github.com/clientIO/joint/blob/b1c5763ae5d1887644a9a0ac1c4845a7edfe2b5a/src/joint.dia.graph.js#L108
They are actually sorted at the end of the batch for performance reasons.

I think that a possible solution is to use `maxZIndex` remove the optimization on cell sort and add the optimization on the paper itself.
https://github.com/clientIO/joint/blob/b1c5763ae5d1887644a9a0ac1c4845a7edfe2b5a/src/joint.dia.paper.js#L309
By checking even for to-front and to-back batches.

What do you think?

src/joint.dia.cell.js
B3rn475: done

kumilingus:
We are not querying `maxZIndex` inside the `to-front`/`to-back` batch.
https://github.com/B3rn475/joint/blob/296192607e361759032d707ced1990730d3c1748/src/joint.dia.cell.js#L201-L209

The situation you have described can happen only if someone starts `to-front` batch manually.
```javascript
// cells are sorted
graph.startBatch('to-front');
el1.toFront({ deep: true });
// cells not sorted
el2.toFront({ deep: true })
// cells not sorted
graph.stopBatch('to-front')
// cells sorted
```

So `el1` and `el2` will be sent to front, so their z-indexes will be larger then all other cells. `el1` and `el2` will have same indexes, which I find acceptable.

Am I missing something?

Moving the logic to the paper might work as well (listening to `batch:stop` and call `sortViews`). We will be sorting the graph after every `z` change though.

There is POC which makes the paper to render/update/sort views in animation frames (render loops) and introducing a new concept of z-index pivots avoiding the need for sorting views.
https://github.com/kumilingus/joint/blob/links/src/joint.dia.paper.js
This is the way we would like to go. Because no such an optimazation (e.g with batches) will be required.

B3rn475:
I do have the exact same situation in one of my projects.

When an event occurs, for graphical reasons I need to `toFront` a set of links which can be relatively big and are not children of the same element.

B3rn475:
I've erroneously added the `dist` folder in the commit.
If you will accept the changes I'll remove it during the squash.

kumilingus:
Hi Carlo, I like it. Can you add `minZIndex()` to the documentation and to the TS definition file? Also `hasActiveBatch()` signature has changed and therefore updated in TS. Thank you.

B3rn475:
Done ;-)

kumilingus:
Nice, Good job! Thanks

B3rn475:
@kumilingus I was doing a double check and I've noticed that I forgot to updated the implementation of `toBack` with the new `minZIndex`.
https://github.com/clientIO/joint/blob/cc25144ed9a05b57cc913b739108d68080fd0083/src/joint.dia.cell.js#L216
Can you fix it? or should I send you another PL?

kumilingus:
Thanks @B3rn475 , I'll fix this now.

PR #824, opened 2018-02-13
Fix to get correct outPorts

JaHicks:
A bug was introduced which removed the "this.get" from before "('outPorts')".  This caused the outPorts to be returned as each letter of 'outPorts' rather than the outPorts themselves.


kumilingus:
Thanks!

PR #870, opened 2018-04-27
Use var to declare vars the first time they are used

ricochet1k:
This fixes several errors of the form: `ReferenceError: assignment to undeclared variable router`


kumilingus:
Thanks @ricochet1k !

PR #871, opened 2018-05-02
Adding removePorts() to Element prototype

brisberg:
## Summary
Adding a new api to the Element prototype for batch removing ports.

## Details
This method will remove all of the specified ports (objects or portId strings) from the ports list before calling .prop().

This means a single .prop() call at the end of the batch which is much faster than removing them individually.

Additionally, it was strange that we had addPort/addPorts but only one removePort in the API.

### DIFF
--- a/src/ports.js
+++ b/src/ports.js
@@ -418,6 +418,25 @@
             return this;
         },
 
+        removePorts: function(ports, opt) {
+
+            var options = opt || {};
+
+            if (ports.length) {
+                options.rewrite = true;
+                var remainingPorts = util.assign([], this.prop('ports/items'));
+                ports.forEach(function(port) {
+                    var id = util.isObject(port) ? port.id : port;
+                    remainingPorts = remainingPorts.filter(function(rPort) {
+                        return rPort.id !== id
+                    });
+                })
+                this.prop('ports/items', remainingPorts, options);
+            }
+
+            return this
+        },
+
         /**
          * @private
          */
--- a/src/ports.js
+++ b/src/ports.js
@@ -418,19 +418,19 @@
             return this;
         },
 
-        removePorts: function(ports, opt) {
+        removePorts: function(portsForRemoval, opt) {
 
             var options = opt || {};
 
-            if (ports.length) {
+            if (portsForRemoval.length) {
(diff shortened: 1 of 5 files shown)


vtalas:
Hi, even the doc is updated, good job. I'm not sure why isn't `removePorts` there from the beginning....

src/ports.js
```diff
@@ -418,6 +418,25 @@
...
+        removePorts: function(ports, opt) {
+
+            var options = opt || {};
+
+            if (ports.length) {
+                options.rewrite = true;
+                var remainingPorts = util.assign([], this.prop('ports/items'));
+                ports.forEach(function(port) {
+                    var id = util.isObject(port) ? port.id : port;
+                    remainingPorts = remainingPorts.filter(function(rPort) {
```
vtalas: wouldn't be even faster if the `filter` weren't here, in every iteration?

brisberg: Do you have psudocode for another way to do it?

I wasn't able to find any better ways to do "array subtraction" in javascript

kumilingus: I would prefer to filter the ports once, but search multiple times.
```javascript
var remainingPorts = currentPorts.filter(cp => !portsForRemoval.some(rp => portsEquals(cp, rp)));
```

kumilingus:
It might be also nice if the method removes all ports when called with no parameter.

1.  The first parameter is an aray
```javascript
removePorts(ports: Array<Element.Port|string>, opt?: Cell.Options): this;
```

2.  Otherwise
```javascript
removePorts(opt?: Cell.Options): this;
```

src/ports.js
brisberg: Thanks, yeah that worked very cleanly.

brisberg:
Awesome, I was considering adding a 'removeAllPorts' call next, but I like folding it into this call to reduce API sprawl.

Pushed up those changes with the alternate call signature, should be ready for rereview.

kumilingus:
Thank you for patience and your work @brisberg.

brisberg:
Thanks, and no problem on the wait. It seems like you guys were busy with the latest release.

PR #872, opened 2018-05-02
Improve joint.dia.graph.addCells() performance.

thorikawa:
## Summary
Improve `joint.dia.graph.addCells()` performance by making a call to backbone's `add()` method once with a collection argument rather than calling it multiple times for each model.

## Detail
I have a graph with >600 cells which takes >75 secs to display with current jointjs implementation. For investigation, I run the performance profiler and I noticed that most of time is consumed in graph's `addCell()` method. With current implementation, backbone's `add()` method is called one by one even if you call graph's `addCells()` method which causes a lot of backbone event triggers and takes very long.

To improve it, I made a modification to graph's `addCells()` method so it calls backbone's `add()` method once with a collection argument.  It prevents backbone event triggers happening multiple times and saves a lot of time.

## Result
After this modification, a graph with >600 cells takes <6 secs.

Before:
perf1

After:
perf2


vtalas:
hi, there are some failing tests, all seems to be `addCells` related. You can find instructions how to run the tests in the readme file. Also please do not modify anything in the `dist` folder. Anyway, your effort is highly appreciated, keep going!

kumilingus:
```
>> basic - async: addCells
>> Message: Test took longer than 5000ms; test timed out.
>> file:///home/travis/build/clientIO/joint/node_modules/qunitjs/qunit/qunit.js:1980:18
>> paper - async paper.addCells() should not throw on non-flat array
>> Message: Test took longer than 5000ms; test timed out.
>> file:///home/travis/build/clientIO/joint/node_modules/qunitjs/qunit/qunit.js:1980:18
>> paper - async paper.addCells() should not throw on non-flat array
>> Message: Expected 2 assertions, but 1 were run
>> file:///home/travis/build/clientIO/joint/test/jointjs/paper.js:65:15
```

1. Adding cells in async mode does not work anymore (might has something to do with `opt.position` flag)
2. `graph.addCells([cell1, [cell2, cell3]])` used to work previously

thorikawa:
It turns out that just updating to v2.1.0 and calling addCells() at once  improved performance a lot. My change still improves a bit, but as the test result said, my change broke `async` option of paper and I didn't come up with better solution, so decided to close it. Thank you for commenting.

kumilingus:
Hi @thorikawa, fair enough. I have created an issue, so we can go back to this later and investigate further.

You might improve the performance by redefining your shapes and links (check out `shapes.standard` definitions for best practices;  new `standard` links without the tools included in the markup but rendered on-hover via `linkTools`)

Thank you for the contribution.

Issue #879, opened 2018-05-09
Paper cell:contextmenu not working on 2.1.2


Thunkar:
I have an Angular app that has been working perfectly fine on version 2.0.1. After updating, this code has stopped working: 

```
this.paper.on('cell:contextmenu', (cellView, evt, x, y) => {
      this.elementsRightClicked.emit();
})
```

The event is never fired, and the browser shows the default context menu. I have also tried element:contextmenu and setting preventContextMenu on the paper options to no avail.
If I capture the event in my component using the decorator @HostListener and run event.preventDefault() the browser's context menu is not displayed.

Thunkar:
The issue only seems to be happening on Linux systems, regardless of browser (which seems pretty strange, but there's that).

I've tried debugging the copy of joint.js that angular bundles with my code and the event is never fired internally (options.preventContextMenu is never checked). I'm currently looking through the refactored paper events code, but have yet to find a clue. Reverting to 2.0.1 is the only way I can make it work.

vtalas:
hi, thank you for reporting this, `contextmenu` event is not fired on paper. Tested in Chrome, Firefox in Ubuntu

kumilingus:
The paper events are undelegated on `pointerdown` event in from `v2.1.0`. This causes the `contextmenu` event not to fire.

Here's a workaround for the time being. It prevents `pointerdown` event from firing when the right mouse button is pressed.

```
paper.options.guard = function(evt) {
  return (evt.type === 'mousedown' && evt.buttons === 2);
};
```

kumilingus:
Fixed in https://github.com/clientIO/joint/pull/889.

PR #889, opened 2018-05-25
dia.Paper: fix contextmenu event (pointerdown event not fired on mous…

kumilingus:
…e right click)

Fix for (#879)


nags178:
can i see the resolution how it was fixed. I am unable to do right click in paper

PR #997, opened 2018-10-29
dia.Paper.Options: add InteractivityOptions as function return type (Declaration files)

jm9e:
TypeScipe declaration files are missing `CellView.InteractivityOptions` for dia.Paper.Options.interactive as possible return type if using a function.


kumilingus:
Thanks!

Issue #957, opened 2018-09-14
layout.DirectedGraph: remove redundant vertices


kumilingus:
1. implement `g.Polyline.prototype.filterRedundant(): Polyline` 
2. use the method 1) in the layout.DirectedGraph for `setVertices()` default
3. use the method 1) in the dia.LinkView for `removeRedundantLinearVertices()`

kumilingus:
Fixed in `v3.1.0`.

PR #1014, opened 2018-11-22
g.Polyline: simplify() (WIP)

zbynekstara:
Resolves issue #957.


zbynekstara:
needs to accept a precision option for points and angle comparison

PR #1045, opened 2019-02-15
Support scientific notation in paths

bendemboski:
SVG supports this, and in some cases, V.normalizePathData() can produce almost-zero numbers that use scientific notation. This means that calling V.toGeometryShape() on a path can result in a V.normalizePathData() call that introduces scientific notation to the path, so without supporting scientific notation in the path parsing, the user can get an `e is not a recognized path segment type` error from deep inside joint code.

### DIFF
--- a/src/geometry.js
+++ b/src/geometry.js
@@ -1718,14 +1718,14 @@ var g = {};
 
         var path = new Path();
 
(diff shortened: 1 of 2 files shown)


kumilingus:
Thanks, Ben!

PR #1051, opened 2019-02-25
Geometry: containsPoint

zbynekstara:


src/geometry.js
```diff
@@ -1370,6 +1380,22 @@ var g = {};
...
+        // Returns `true` if the point lies on the line.
+        containsPoint: function(p) {
+
+            var start = this.start;
+            var end = this.end;
+
+            if (start.cross(p, end) !== 0) return false;
+            // cross product of 0 indicates that this line and the vector to `p` are collinear
+
+            if ((new g.Line(start, p)).length() > this.length()) return false;
```
kumilingus: Let's store `this.length()` into a variable here as it's calculated twice.

PR #1072, opened 2019-05-03
Added util ailas for _.forEach and using it to iterate over arrays

brisberg:
tl;dr util.forIn operating on an object is optimized away when using the Clojure compiler

--

The lodash utilities forIn and forEach have slightly different purposes; forIn iterates over an objects properties and forEach iterates over an array or iterable. 

While forIn technically works for iterating over arrays, it is not ideal. In these specific cases, this.getOutboundEdges() will return an array of edges, not an object. 

These were changed from _.each into util.forIn recently in this [commit](https://github.com/clientIO/joint/commit/af22c84e838b913f09299cbbe2fa02fbfc6c4421). I believe this was an oversight since it is operating on an array not an object.

This isn't an issue for most users as it does in fact 'work'. However, my project is using the Clojure compiler to optimize our bundle (including jointjs) and the Clojure compiler is 'smart' enough to notice this mismatch. It decides that this loop will never produce useful output and optimizes most of the function away, thus always returning [] for the list of connected links.

I believe this is the simplest fix for the issue, and it works exactly the same for uncompiled users.

### DIFF
--- a/src/graph.js
+++ b/src/graph.js
@@ -445,7 +445,7 @@ export const Graph = Backbone.Model.extend({
         var edges = {};
 
         if (outbound) {
-            util.forIn(this.getOutboundEdges(model.id), function(exists, edge) {
+            util.forEach(this.getOutboundEdges(model.id), function(exists, edge) {
                 if (!edges[edge]) {
                     links.push(this.getCell(edge));
                     edges[edge] = true;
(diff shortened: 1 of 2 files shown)


kumilingus:
> These were changed from _.each into util.forIn recently in this commit. I believe this was an oversight since it is operating on an array not an object.

Are you sure? I can see that `graph.getInbountEdges()` and `graph.getOutboundEdges()` return an object `{}` (not an array `[]`).

PR #1073, opened 2019-05-03
Updated typings and initializes _V cache on render of links with MarkupJSON

brisberg:
Links can be specified using the MarkupJSON format instead of a xml string. This does produce faster rendering since it doesn't need to re-parse the string every time which is great.

However, I noticed that links built this way do not get moved by directedGraph.layoutGraph(). Digging into it it seems to be cause the renderJSON function doesn't actually build the _V vector cache of the svg elements, which directedGraph uses to layout the edges.

This should emulate the behavior of renderMarkup. Though the only difference is that the _V cache will contain one extra element for 'root'.

I also updated the typings so that MarkupJSON can be used from typescript

### DIFF
--- a/src/joint.dia.link.js
+++ b/src/joint.dia.link.js
@@ -761,6 +761,7 @@ joint.dia.LinkView = joint.dia.CellView.extend({
         var doc = this.parseDOMJSON(markup, this.el);
         // Selectors
         this.selectors = doc.selectors;
+        Object.keys(this.selectors).forEach((key) => this._V[key] = V(this.selectors[key]));
         // Fragment
         this.vel.append(doc.fragment);
     },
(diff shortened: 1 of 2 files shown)


kumilingus:
Did you define a custom link using the MarkupJSON and trying to mimic the original XML string? Keeping all the classNames? Could you please send me the link definition?

The `_V` property is sort of deprecated and it's there only for the backwards compatibility of the default `joint.dia.Link`. This old link contains the tools, arrowheads and vertices in the markup. This is slow to render and update and it is visible only when the user hovers the link. It's suggested to switch to `joint.shapes.standard.Link` and use `joint.dia.LinkTools` that add these extra DOM elements on demand.

I am not saying this PR is wrong. I want to know more about your use-case first. Perhaps we could also change the `joint.dia.Link` default markup to the JSON.

PR #1077, opened 2019-05-15
Fix normalizing paths with zero-radius arcs

bendemboski:
The path normalizing code was not checking the radii for zero, and was ending up with NaN in path instructions in this case. W3C (https://www.w3.org/TR/SVG/paths.html#ArcOutOfRangeParameters) instructs user agents to treat it as a lineto in this case.


kumilingus:
Well done, @bendemboski, thanks.

PR #1153, opened 2019-07-31
Add a browser key to point to dist

deanohyeah:
By default webpack look for browser, module, main. It is desirable to use the dist for web builds.

Currently, jointjs blow up in ie 11 because it's using the mjs module entry point.


kumilingus:
Nice! Thanks.

PR #1184, opened 2019-09-11
g.Polyline: simplify() (remove redundant vertices)

zbynekstara:
Resolves issue #957.


zbynekstara:
Should also accept a precision option for points and angle comparison.

zbynekstara:
Applied also at the places it counts:
- `linkView.removeRedundantVertices`
- DirectedGraph demo

Issue #1168, opened 2019-08-28
pointerup event on link label does not fire


ArthurWietzorek:
I am using Rappid 3.0.1 to develop a typescript angular application and am facing a problem i can't seem to solve. I am trying to use the TextEditor to inline edit the label of a link, but it won't work if i click on the label.

I added the following code to one of my components to observe the events triggered on my paper:

```
    this.paper.on('cell:pointerdown', () => {
      console.log('cell:pointerdown');
    });
    this.paper.on('cell:pointerup', () => {
      console.log('cell:pointerup');
    });
    this.paper.on('element:pointerdown', () => {
      console.log('element:pointerdown');
    });
    this.paper.on('element:pointerup', () => {
      console.log('element:pointerup');
    });
    this.paper.on('link:pointerdown', () => {
      console.log('link:pointerdown');
    });
    this.paper.on('link:pointerup', () => {
      console.log('link:pointerup');
    });
```

The following happens:
![missingPointerup](https://user-images.githubusercontent.com/41048767/63856629-65b8b700-c9a2-11e9-852b-476e6f8bfe1d.gif)

For some reason, the pointerup event does not get triggered on the link label. Do labels have their own event names? What other possible mistakes could i have done using Rappid?

kumilingus:
It's a bug. It happens, when `moveLabel` interactivity option is set to `false`.

You might override `dragLabelStart` method of the LinkView in the meantime.
```js
joint.dia.Linkiew.prototype.dragLabelStart = function(evt, x, y) {

        if (!this.can('labelMove')) {
            this.eventData(evt, { stopPropagation: true });
            this.paper.delegateDragEvents(this, evt.data);
            return;
        }

        var labelNode = evt.currentTarget;
        var labelIdx = parseInt(labelNode.getAttribute('label-idx'), 10);

        var positionAngle = this._getLabelPositionAngle(labelIdx);
        var labelPositionArgs = this._getLabelPositionArgs(labelIdx);
        var defaultLabelPositionArgs = this._getDefaultLabelPositionArgs();
        var positionArgs = this._mergeLabelPositionArgs(labelPositionArgs, defaultLabelPositionArgs);

        this.eventData(evt, {
            action: 'label-move',
            labelIdx: labelIdx,
            positionAngle: positionAngle,
            positionArgs: positionArgs,
            stopPropagation: true
        });

        this.paper.delegateDragEvents(this, evt.data);
    };
```
[...]

PR #1186, opened 2019-09-18
dia.LinkView: fix missing pointerup event for non-interactive label

kumilingus:


kumilingus:
Fix https://github.com/clientIO/joint/issues/1168.

PR #1195, opened 2019-10-05
types: use proper jquery event.

### LINKED ISSUE #1194: Types: incorrenct jquery interface for the event object
I think that the interface used for the event object in the typing of jointjs use the wrong jquery interface: `JQuery.Event` instead of `JQuery.TriggeredEvent`

reference stackoverflow https://stackoverflow.com/questions/55931254/typescript-jquery-event-in-handler-has-not-properties-target-and-current

The `JQuery.Event` interface does not have the `target`, `data`, and some other useful properties, that make working with it annoying

I think the problem lays mainly in the naming choice in the jquery typing that is confusing.

I can make a PR if the type is indeed not correct

CaselIT:
Added an alias type `dia.Event`, using `JQuery.TriggeredEvent` instead of `JQuery.Event` for the callbacks.

Fixes #1194

### DIFF
--- a/types/joint.d.ts
+++ b/types/joint.d.ts
@@ -6,6 +6,8 @@ export namespace config {
 
 export namespace dia {
 
+    type Event = JQuery.TriggeredEvent;
+
     type Point = g.PlainPoint;
 
     type BBox = g.PlainRect;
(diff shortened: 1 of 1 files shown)


kumilingus:
Just checking - is it ok in TS to refer to `Event` as `dia.Event` inside the `dia` namespace?

CaselIT:
It seems so, it does not give any error. I've also tried to run `grunt dist` and replace the generated typing in an project that uses it and it seems to be working.

But I can remove the `dia` in the namespece if you prefer

kumilingus:
Thanks, it works fine.

PR #1238, opened 2020-01-03
Geometry: add parallel() to Line, add serialize() to Point and Line

kumilingus:


zbynekstara:
`parallel` could use `offset` with trigoniometry functions for coords - avoids multiple cos/sin calculations

PR #1325, opened 2020-06-20
dia.Paper: improve markers, gradients and patterns

kumilingus:
- SVGMarkers can be now easily built from more than one element
- Add higher-level API to define SVGPatterns
- Add Typescript definition for Filters / Markers / Gradients
- Replace `gradients` with `fills` demo
image


demo/shapes/fillings.html
```diff
@@ -2,29 +2,22 @@
 <html>
     <head>
         <meta charset="utf8"/>
-
-        <title>Gradients</title>
-
+        <title>Fillings</title>
```
jamesgeorgewilliams, line 5: I think I would change the `<title>` to `Fills` and, `<script>` and filenames to `./src/fills.js + demo/shapes/fills.html`. 'Fillings' is a correct word, but not so usual in this context I think. I know `fill` and `gradient` are different when talking about the specific thing, and maybe you don't want to use the word 'fill', but I think in a general sense, a 'gradient' can be considered a type of 'fill' too. So `fills` makes sense I think. The rest of the docs look good.

docs/src/joint/api/dia/Paper/prototype/defineGradient.html
```diff
@@ -1,27 +1,61 @@
...
+    </tr>
+    <tr>
+        <td>type</td>
+        <td><code>'linearGradient'</code> | <code>'radialGradient'</code></td>
+        <td>The <a target="_blank" href=https://developer.mozilla.org/en-US/docs/Web/SVG/Element/linearGradient>linearGradient</a> or <a target="_blank" href=https://developer.mozilla.org/en-US/docs/Web/SVG/Element/radialGradient>radialGradient</a> type of the gradient.</td>
+    </tr>
+    <tr>
+        <td>id</td>
+        <td>string</td>
+        <td> <i>(optional)</i> - A unique identifier of the gradient. The <i>id</i> is generated if none provided.</td>
```
jamesgeorgewilliams, line 19: There is an extra space here. The same on:
line `15` in `docs/src/joint/api/dia/Paper/prototype/defineMarker.html ` +
line `14` in `docs/src/joint/api/dia/Paper/prototype/definePattern.html`

PR #1348, opened 2020-08-19
g.Path: add toSubpathsSegments()

kumilingus:


zbynekstara:
Alternative PR: #1350

Issue #1356, opened 2020-09-12
Connector jumpover leading cell selection slow.


asahu132:
Hi Team,
I am facing an issue while using **jumpover** connector it is attaching one more extra listner on event **cell:pointerup** which is making over app slow to select a cell when the graph size gets increase.

abhisheksahu132:
@kumilingus are you guys working on this performance issue?
 This issue has been fixed in #1357

kumilingus:
Hi @asahu132, I haven't started yet. We'll definitely look into it before the next release. At this point I am not sure if listening to `batch:stop` is the best solution (it seems better than `cell:pointerup` though 👍 ).

kumilingus:
Fixed by https://github.com/clientIO/joint/commit/42d7f86a274fbabcabda87e8bcd4c1988ef54c71.

PR #1357, opened 2020-09-13
change cell:pointerUp event to batch:stop for jumpoverUpdate.

abhisheksahu132:
[Fixed issue](https://github.com/clientIO/joint/issues/1356)

### DIFF
--- a/src/connectors/jumpover.mjs
+++ b/src/connectors/jumpover.mjs
@@ -47,7 +47,7 @@ function setupUpdating(jumpOverLinkView) {
     // first time setup for this paper
     if (updateList == null) {
         updateList = jumpOverLinkView.paper._jumpOverUpdateList = [];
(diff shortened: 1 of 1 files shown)


src/connectors/jumpover.mjs
```diff
@@ -47,7 +47,7 @@ function setupUpdating(jumpOverLinkView) {
     // first time setup for this paper
     if (updateList == null) {
         updateList = jumpOverLinkView.paper._jumpOverUpdateList = [];
-        jumpOverLinkView.paper.on('cell:pointerup', updateJumpOver);
+        jumpOverLinkView.paper.on('batch:stop', updateJumpOver);
```
kumilingus, line 50: `"batch:stop"` is an event on `graph` not on `paper`

kumilingus:
Fixed in https://github.com/clientIO/joint/pull/1393.

PR #1375, opened 2020-10-04
Adjust center position of new scale wrt previous scale, not 1.

avshenuk:
Tested locally and this now works correctly. 
A simple example: if you're already on scale 2, and trying to set the scale to 2 again, simply glancing through the previous code, you can see that it will move the origin when it shouldn't at all.

### DIFF
--- a/src/dia/Paper.mjs
+++ b/src/dia/Paper.mjs
@@ -1554,11 +1554,12 @@ export const Paper = View.extend({
             oy = 0;
         }
 
+        var scale = this.scale();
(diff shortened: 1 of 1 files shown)


kumilingus:
Hi @avshenuk, thank you! This looks good. I just need to test whether it is a breaking change or not.

kumilingus:
Sorry for the delay. It is a breaking fix and it was addressed here: https://github.com/clientIO/joint/pull/2495
Thanks for pointing that out.

PR #1394, opened 2020-11-18
Support inserting port at position

UXabre:
I've added the possibility to insert ports at specific indexes. This is important if we want to allow dynamic port additions to be possible while having a logically sorted list of ports.

PoC

Hope you guys like it!

### DIFF
--- a/src/dia/ports.mjs
+++ b/src/dia/ports.mjs
@@ -335,6 +335,26 @@ export const elementPortPrototype = {
     },
 
     /**
+     * @param {string|Port|number} before
+     * @param {object} port
+     * @param {object} [opt]
+     * @returns {joint.dia.Element}
+     */
+    insertPort: function(before, port, opt) {
+        const index = (typeof before === 'number') ? before : this.getPortIndex(before);
+
+        if (!util.isObject(port) || Array.isArray(port)) {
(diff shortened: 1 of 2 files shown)


src/dia/ports.mjs
```diff
@@ -334,6 +334,26 @@ export const elementPortPrototype = {
...
+     * @param {string|Port|number} before
+     * @param {object} port
+     * @param {object} [opt]
+     * @returns {joint.dia.Element}
+     */
+    insertPort: function(before, port, opt) {
+        const index = (typeof before === 'number') ? before : this.getPortIndex(before);
+
+        if (!util.isObject(port) || Array.isArray(port)) {
+            throw new Error('Element: addPort requires an object.');
```
kumilingus, line 347: Could you please change this to `dia.Element: insertPort requires an object.`?

kumilingus:
Hi @UXabre , yes, we do like it :) I've added a comment. Once this is changed I'll merge this.

src/dia/ports.mjs
UXabre: Hi @kumilingus , oh, copy-paste error for sure :-( I'll change this ASAP, should've seen this myself. THanks for the review

PR #1420, opened 2021-02-05
feat(): add vue-html demo

dawidKrynicz:


IanSietro:
Alternated by https://github.com/clientIO/joint/pull/1435 for Vue 3

PR #1440, opened 2021-03-02
elementTools.Remove: fix example in documentation

dcecile:
The element tool will pass a link view to the remove action, right?


kumilingus:
It is an `elementView` passed to the callback. It was a leftover error after we copied the code from the `linkTools.Remove`. Thanks!

dcecile:
No problem 🙂

Issue #1469, opened 2021-05-13
Remove support for IE9 polyfill for XHR object


RuslanZavacky:
This polyfill seems to be quite old, as library itself states that it support IE11+.

Also, due to this polyfill, there is a problem to use this library in the same time with Pretender.js. Pretender is "faking" the XMLHttp object, so it can intercept requests, and this polyfill can't detect that, as response is not on the object of pretender. [Reference 1 - FakeXmlHttpRequest](https://github.com/pretenderjs/FakeXMLHttpRequest/blob/master/src/fake-xml-http-request.js) [Reference 2 - Pretender.js](https://github.com/pretenderjs/pretender/blob/ddca21671e3e7c69b0be673eb3cd66007931a379/src/pretender.ts)

I'll prepare a PR to remove this polyfill, is there any objections to do so?

Thank you in advance!

kumilingus:
Removed in https://github.com/clientIO/joint/pull/1470.

PR #1470, opened 2021-05-13
fix(GH-1469): Remove polyfill for XMLHttpRequest support of IE9

RuslanZavacky:
As library aims to support IE11 and up, this polyfill seems to be not needed anymore. As well as this polyfill breaks the use of Pretender io and joint.js library.


RuslanZavacky:
Thank you @vtalas for such quick turn around. Do you know when you can release a new version with this change?

PR #1480, opened 2021-06-07
dia.Graph: fix order of cells in getElements() and getLinks()

kumilingus:
- Fixes an issue with embedding mode, which relies on `getElements()` returning the elements in the graph collection order (the elements being sorted by their `z`). The current solution returned the elements (links) in unspecified order (based on object properties order).
- Improve performance for typical graphs where number of elements is similar to number of links (https://jsfiddle.net/kumilingus/L48xyr1f/)


jamesgeorgewilliams (commented):
Looks good. I haven't used Qunit before, but the syntax seems similar to Jest. I will take a look at Qunit in more detail.

jamesgeorgewilliams (approved):

PR #1489, opened 2021-06-28
 dia.Paper: add validateUnembedding() option

kumilingus:
Add missing validation callback for `embeddingMode` to check if an element can become a new root (to be un-embedded from a parent and become an element without parent).

If invalid -> the position and parent-child relationship is reverted to the state before user interaction.

Rappid Stencil (or an element palette) should take this option into account when an element is dropped from the stencil and the `embeddingMode` is on.

### DIFF
--- a/docs/src/joint/api/dia/Paper/prototype/options/validateUnembedding.html
+++ b/docs/src/joint/api/dia/Paper/prototype/options/validateUnembedding.html
@@ -1,0 +1,10 @@
+<code>validateUnembedding</code> - a function that allows you to control which elements can be unembedded when the paper is put into the <code>embeddingMode</code> <i>(see above)</i>.
+<br>
+The function signature is <code>function(childView)</code> and should return
+<code>true</code> if the <code>childView</code> element can be unembedded from the parent and become a new root.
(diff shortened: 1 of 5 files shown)


docs/src/joint/api/dia/Paper/prototype/options/validateUnembedding.html
```diff
@@ -0,0 +1,10 @@
+<code>validateUnembedding</code> - a function that allows you to control which elements can be unembedded when the paper is put into the <code>embeddingMode</code> <i>(see above)</i>.
+<br>
+The function signature is <code>function(childView)</code> and should return
+<code>true</code> if the <code>childView</code> element can be unembedded from the parent and become a new root.
+<br>
+By default, all elements can become roots (the function returns <code>true</code> no matter what).
+<br>
+The callback is called at the end of drag & drop action and if <code>false</code> is returned, the position of the element, as well the parent reference is reverted to the values before dragging.
```
jamesgeorgewilliams, line 8: Looks good. In this sentence, you can change to ", as well as the parent reference".

PR #1512, opened 2021-09-06
dia.CellView: update typings for ES6 class extension

kumilingus:
Fix various issues one would face when inheriting `cellView` classes in Typescript.

```ts
class MyView extends dia.ElementView {
  presentationAttributes() {
     return { /* */ }
  }

  render() {
    /* ... */
    this.updateTransformation();
  }
}
```

### DIFF
--- a/types/joint.d.ts
+++ b/types/joint.d.ts
@@ -607,6 +607,15 @@ export namespace dia {
 
         type FlagLabel = string | string[];
         type PresentationAttributes = { [key: string]: FlagLabel };
+
+        type NodeData = { [key: string]: any };
+
+        type NodeMetrics = {
+            data: NodeData;
+            boundingRect: g.Rect;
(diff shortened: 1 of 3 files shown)


types/joint.d.ts
```diff
@@ -615,9 +624,9 @@ export namespace dia {
 
         paper: Paper | null;
 
-        initFlag: CellView.FlagLabel;
+        initFlag(): CellView.FlagLabel;
 
-        presentationAttributes: CellView.PresentationAttributes;
+        presentationAttributes(): CellView.PresentationAttributes
```
jamesgeorgewilliams, line 629: Looks good, I should have remembered to change `initFLag` and `presentationAttributes`. I think there is a semi-colon missing here. Part of the build process has failed on this PR for some reason.

PR #1519, opened 2021-09-10
linkTools, elementTools: add Connect button

kumilingus:


src/dia/CellView.mjs
```diff
@@ -398,6 +398,23 @@ export const CellView = View.extend({
...
+        var paper = this.paper;
+        var graph = paper.model;
+
+        var link = paper.getDefaultLink(this, magnet);
+        link.set({
+            source: this.getLinkEnd(magnet, x, y, link, 'source'),
+            target: { x: x, y: y }
+        }).addTo(graph, {
+            async: false,
+            ui: true
```
jamesgeorgewilliams, line 412: What does `ui: true` option do again? That is is present in the model or not? I don't know if we have it documented.

kumilingus: Each time the user makes a change to a model from the UI we pass `ui: true` flag. It is only to say if a given change was made by the user or not.

```js
graph.on('change:position', (element, position, opt) => {
  if (opt.ui) {
    // the user drag & drop the element
  } else {
   // the developer shifted the element e.g. element.translate(10, 10)
  }
});

jamesgeorgewilliams: Okay, thanks

PR #1523, opened 2021-09-13
dia.Cell, dia.Graph: better use of generic class types

kumilingus:


types/joint.d.ts
```diff
@@ -372,9 +390,7 @@ export namespace dia {
         }
     }
 
-    class Element extends Cell {
-
-        constructor(attributes?: Element.Attributes, opt?: Graph.Options);
+    class Element<A = Element.Attributes, S = dia.ModelSetOptions> extends Cell<A,S> {
```
jamesgeorgewilliams, line 393: I think there is a space missing here, and the same on line :516 `Cell<A, S>`

PR #1543, opened 2021-10-04
dia.attributes: support nested calc() functions

kumilingus:


docs/src/joint/api/dia/attributes/intro.html
```diff
@@ -98,7 +98,6 @@ <h5 class="calc-header">Syntax:</h5>
     </li>
     <li>
         <span class="calc-part">Addition or Subtraction</span> is an optional floating number added or subtracted from the variable. It's a <code>+</code> or <code>-</code> symbol followed by a number.
-        <pre><code>+ 5</code></pre>
```
jamesgeorgewilliams, line 101: Did you mean to delete this line, or maybe I don't understand why it is deleted?

kumilingus: Good catch!

PR #1569, opened 2021-11-09
connectors.curve: a new curved connector

Geliogabalus:
Improved default smooth connector behaviour


Geliogabalus:
Created direction option with modes:
'auto' - changes curve depending on closest side (via linkView.sourceBBox.sideNearestToPoint(linkView.getEndConnectionPoint('source'))). Works similar as default but has better behaviour near change points (didn't gets odd look). I suggest to change default render to same as in auto.
'horizontal' - always renders connection as horizontal. 
'vertical' - always renders connection as vertical.
Horizontal and vertical work better with anchors and 'connectionPoint: { name: 'anchor' }'

docs/src/joint/api/connectors/smooth.html
```diff
@@ -1,8 +1,32 @@
...
+<p>The <code>'smooth'</code> connector interpolates route points using a cubic bezier curve. Without a route it uses options to render the link.</p>
 
-<p>Example:</p>
+<p>Available options:</p>
+<table>
+    <tr>
+        <th>direction</th>
+        <td><i>string</i></td>
+        <td>Direction of the connector's curve. 
+            <p>There is exported <code>Directions</code> enum object that provides values for this option</p>
```
jamesgeorgewilliams, line 9: Looks good. For this sentence, you can say "There is **an** exported...".

docs/src/joint/api/connectors/smooth.html
```diff
@@ -1,8 +1,32 @@
-<p>The <code>'smooth'</code> connector interpolates route points using a cubic bezier curve.</p>
+<p>The <code>'smooth'</code> connector interpolates route points using a cubic bezier curve. Without a route it uses options to render the link.</p>
```
jamesgeorgewilliams, line 1: Since options can be considered if a route is provided too, you could say something like _"...curve. Additional options can be provided to render the link"_ or _"...curve. Additional options can be provided to compliment link rendering."_

docs/src/joint/api/connectors/smooth.html
```diff
@@ -1,8 +1,32 @@
...
+                <li><b>Directions.HORIZONTAL</b> - fixes the direction in the horizontal orientation</li>
+                <li><b>Directions.VERTICAL</b> - fixes the direction in the vertical orientation</li>
+                <li><b>Directions.LEGACY</b> - enables a legacy connection render algorithm</li>
+            </ul>
+        </td>
+    </tr>
+    <tr>
+        <th>minOffset</th>
+        <td><i>number</i></td>
+        <td>Minimal curve point offset in case of the source and target proximity. Default is <code>100</code>.</td>
```
jamesgeorgewilliams, line 21: "...in case of source and target proximity". (without 'the')

"The default is... " (with 'the')

docs/src/joint/api/connectors/smooth.html
Geliogabalus: In the current implementation options don't matter when route points exist. I inherited this from previous implementation.

jamesgeorgewilliams: Okay. I think it is fine so.

PR #1575, opened 2021-11-15
Geometry: intersection of  polyline with other objects

kumilingus:


docs/src/geometry/api/g/intersection/exists.html
```diff
@@ -0,0 +1,26 @@
...
+    <li><a href="#g.Rect">Rect</a></li>
+    <li><a href="#g.Polyline">Polyline</a></li>
+    <li><a href="#g.Polygon">Polygon</a></li>
+    <li><a href="#g.Path">Path</a></li>
+</ul>
+
+<pre><code>
+g.intersection.exists(new g.Rect(0, 0, 10, 10), new g.Line({ x: 5, y: 5 }, { x: 30, y: 30 })); // true</code></pre>
+
+Using the the shape options is valid for <a href="#g.Path">Path</a> only. The path is internally convert into polylines with <a href="#g.Path.prototype.toPolylines">toPolylines()</a> method. The options control the conversion of the path.
```
jamesgeorgewilliams, line 15: Looks good. The code comments are very clear.

"Using **the** shape options..." (1 'the').

"The path is internally **converted** into polylines with **the** ..."

PR #1581, opened 2021-11-23
dia.attributes: add x and y properties to calc() expression

kumilingus:
Add `x` and `y` properties to `calc()` in order to support this use case

```js
element.attr('myTextBackground', {
    ref: 'myText',
    refX: -5,
    refY: -5,
    width: 'calc(w + 10)',
    height: 'calc(h + 10)',
 });
```

without a need of using `refX` and `refY` attributes.
```js
element.attr('myTextBackground', {
    ref: 'myText',
    x: 'calc(x - 5)',
    y: 'calc(y - 5)',
    width: 'calc(w + 10)',
    height: 'calc(h + 10)',
 });
 ```


rodritorrico (approved):
Looks good!

PR #1593, opened 2022-01-05
dia.Paper: migrate mousewheel/DOMMouseScroll to wheel

stuhlmeier:
Currently, the mousewheel element event is delegated to the DOM events `mousewheel` and `DOMMouseScroll`.
However, both of these are deprecated/non-standard ([1](https://developer.mozilla.org/en-US/docs/Web/API/Element/mousewheel_event), [2](https://developer.mozilla.org/en-US/docs/Web/API/Element/DOMMouseScroll_event)) and have been replaced by [`wheel`](https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event).

While Chromium will correctly provide a `WheelEvent` regardless (and thus `event.originalEvent` will be a `WheelEvent`), on Gecko/Firefox the event callback will receive a `DOMMouseScroll` instead.
This is also significantly less useful because `DOMMouseScroll` only contains one scroll value (`detail`), so there is no information available about x/y scroll amounts, and it's impossible to know which scroll direction triggered the event.

This change makes it so that the mousewheel event is only triggered by `wheel` mouse events instead.

---

One thing that might also need to be addressed is that `WheelEvent.wheelDelta` is _also_ deprecated; the proper alternative would be to provide both `deltaY` and `deltaX` to the mousewheel event handler.

However since this would be a fairly significant change in functionality, due to an extra callback parameter, I've left that as is for now because that information is still accessible via `Event.originalEvent`.


kumilingus:
Thanks for the PR and the explanation.

PR #1597, opened 2022-01-17
types: add Port and PortLabel layout types

### LINKED ISSUE #1438: joint.layout.Port types missing
Per the documentation there's a namespace for `joint.layout.Port` and `joint.layout.PortLabel`. But in typescript, those do not exist. Any ideas?

I'm trying to do the following: 
```javascript
{
  position: joint.layout.PortLabel.left
}

```
The documentation that says this namespace exists can be found here: https://resources.jointjs.com/docs/jointjs/v3.3/joint.html#dia.Element.ports

jamesgeorgewilliams:
Fixes #1438.

### DIFF
--- a/types/joint.d.ts
+++ b/types/joint.d.ts
@@ -3144,7 +3144,7 @@ export namespace layout {
             compensateRotation?: boolean;
         }
 
-        type PortLayoutFunction = (ports: Array<Object>, elBBox: g.Rect, opt: Options) => Array<g.Point>;
+        type LayoutFunction = (ports: Array<Object>, elBBox: g.Rect, opt: Options) => Array<g.Point>;
 
         export var absolute: PortLayoutFunction;
         export var fn: PortLayoutFunction;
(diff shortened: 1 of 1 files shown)


types/joint.d.ts
```diff
@@ -3041,6 +3041,66 @@ export namespace layout {
 
         export function fromGraphLib(glGraph: any, opt?: { [key: string]: any }): dia.Graph;
     }
+
+    export namespace Port {
+
+        interface PortOptions {
```
kumilingus, line 3047: Just `Options` will do. It is inside of the `Port` namespace. The same for `PortLabelOptions`.

types/joint.d.ts
```diff
@@ -3041,6 +3041,66 @@ export namespace layout {
...
+            dy?: number;
+            angle?: number;
+            start?: g.Point;
+            end?: g.Point;
+            startAngle?: number;
+            step?: number;
+            compensateRotation?: boolean;
+        }
+
+        type PortLayoutFunction = (ports: Array<Object>, elBBox: g.Rect, opt?: PortOptions) => Array<g.Point>;
```
kumilingus, line 3060: `opt` here is not optional. JointJS always pass the option object. Same for LabelLayout callback.

PR #1645, opened 2022-04-07
connectors.Curve: rotate option

Geliogabalus:
Add `rotate` option to `connector.Curve`.
```js
el.set('angle', 45);
el.set('connector', {
    name: 'curve',
    args: {
        rotate: true // take the element's angle into account
    }
});
```

### DIFF
--- a/src/connectors/curve.mjs
+++ b/src/connectors/curve.mjs
@@ -102,7 +102,7 @@ function getHorizontalSourceDirection(linkView, route, options) {
 
     let sourceSide;
     let rotation;
-    if (!sourceBBox.width || !sourceBBox.height) {
+    if (!linkView.sourceView) {
         if (sourceBBox.x > route[1].x)
             sourceSide = 'right';
         else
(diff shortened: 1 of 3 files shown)


src/connectors/curve.mjs
```diff
@@ -100,132 +101,245 @@ function getHorizontalSourceDirection(linkView, route, options) {
...
     let sourceSide;
+    let rotation;
     if (!sourceBBox.width || !sourceBBox.height) {
         if (sourceBBox.x > route[1].x)
             sourceSide = 'right';
         else
             sourceSide = 'left';
     } else {
-        sourceSide = sourceBBox.sideNearestToPoint(route[0]);
+        if (options.rotate) {
```
kumilingus, line 111: Can we perhaps execute this branch only if `options.rotate === true` and `linkView.sourceView.model.angle() !== 0`?

The elements being not rotated is the most common case and we should optimize the code for this.

kumilingus, line 111: Can we perhaps execute this branch only if `options.rotate === true` and `linkView.sourceView.model.angle() !== 0`?

The elements being not rotated is the most common case and we should optimize the code for this.

PR #1658, opened 2022-05-03
util.getRectPoint: add utility to get a point on a rect from given keyword

kumilingus:
Unify the way to search for points on rectangles using keywords in different parts of the library.


Geliogabalus:
Looks nice

Geliogabalus (approved):

PR #1723, opened 2022-07-07
types: fix Paper.Options el type

danielpza:
reference: https://resources.jointjs.com/docs/jointjs/v3.5/joint.html#dia.Paper.constructor

```tsx
var graph = new joint.dia.Graph
var paper = new joint.dia.Paper({
    el: $('#paper'),
    width: 600,
    height: 400,
    gridSize: 10,
    model: graph
});
```

previously this example showed an error with typescript about the property `el` not being defined.


danielpza:
The error is a bit different closing for now

PR #1729, opened 2022-07-11
types: fix cell.prop object types

danielpza:
Fixes

```ts
const cell = new dia.Cell<{foo: string, bar: string}>({foo: 'foo', bar: 'bar'})

cell.prop({foo: 'foo2'}) // this would fail because it was asking for the bar property
```


kumilingus:
Thanks!

PR #1730, opened 2022-07-11
types: allow passing generic attributes to links under shapes.standard

danielpza:
dia.Link currently allows this but shapes.standard.Link does not:

```ts
interface Attributes {
  foo: string;
}

class CustomLink extends dia.Link<Attributes> {} // dia.Link is generic

class CustomLink2 extends shapes.standard.Link<Attributes> {} // this is not possible with current types

const link = new CustomLink({
  // nice autocomplete here in the editor for `foo` variable
  foo: "foo",
});

const link2 = new CustomLink2({
  // no auto completion and no warnings that a number was passed instead of a string
  foo: 3,
});
```

### DIFF
--- a/types/joint.d.ts
+++ b/types/joint.d.ts
@@ -2153,7 +2153,7 @@ export namespace shapes {
 
         type LinkAttributes = dia.Link.GenericAttributes<LinkSelectors>;
 
-        class Link extends dia.Link<LinkAttributes> {
+        class Link<A = dia.Link.Attributes> extends dia.Link<A> {
         }
 
         interface DoubleLinkSelectors extends dia.Cell.Selectors {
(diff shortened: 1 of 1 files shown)


types/joint.d.ts
```diff
@@ -2153,7 +2153,7 @@ export namespace shapes {
 
         type LinkAttributes = dia.Link.GenericAttributes<LinkSelectors>;
 
-        class Link extends dia.Link<LinkAttributes> {
+        class Link<A = dia.Link.Attributes> extends dia.Link<A> {
```
kumilingus, line 2156: Have you changed the default from `dia.Link.GenericAttributes<LinkSelectors>` to `dia.Link.Attributes`?

danielpza: oh, yeah, it was a mistake, I'll correct it

danielpza: should be good now, good catch

PR #1731, opened 2022-07-12
types: allow passing generic attributes to elements

danielpza:
- similar to https://github.com/clientIO/joint/pull/1730


kumilingus:
No issues at all. Thanks!

PR #1788, opened 2022-08-23
dia.LinkView: snap arrowhead to the anchor and vertices

Geliogabalus:
Add `snapLinksSelf` paper option to force a dragged link end to snap to the `x` or `y` coordinate of the closest point on the current link (one of the ends and vertices) if the distance to the point on one axis is lesser than the specified radius.


src/dia/LinkView.mjs
```diff
@@ -41,6 +41,7 @@ export const LinkView = CellView.extend({
         linkToolsOffset: 40,
         doubleLinkToolsOffset: 65,
         sampleInterval: 50,
+        snap: null, // { distance: number } | null
```
kumilingus, line 44: leftover?

PR #1799, opened 2022-08-31
Revert "types: allow passing generic attributes to elements"

kumilingus:
Reverts clientIO/joint#1731

### DIFF
--- a/types/joint.d.ts
+++ b/types/joint.d.ts
@@ -1996,7 +1996,7 @@ export namespace shapes {
 
         type RectangleAttributes = dia.Element.GenericAttributes<RectangleSelectors>;
 
-        class Rectangle<A = RectangleAttributes> extends dia.Element<A> {
+        class Rectangle extends dia.Element<RectangleAttributes> {
         }
 
         interface CircleSelectors extends dia.Cell.Selectors {
@@ -2007,7 +2007,7 @@ export namespace shapes {
 
         type CircleAttributes = dia.Element.GenericAttributes<CircleSelectors>;
 
-        class Circle<A = CircleAttributes> extends dia.Element<A> {
+        class Circle extends dia.Element<CircleAttributes> {
         }
 
         interface EllipseSelectors extends dia.Cell.Selectors {
@@ -2018,7 +2018,7 @@ export namespace shapes {
 
         type EllipseAttributes = dia.Element.GenericAttributes<EllipseSelectors>;
 
-        class Ellipse<A = EllipseAttributes> extends dia.Element<A> {
+        class Ellipse extends dia.Element<EllipseAttributes> {
         }
 
         interface PathSelectors  extends dia.Cell.Selectors {
@@ -2029,7 +2029,7 @@ export namespace shapes {
 
         type PathAttributes = dia.Element.GenericAttributes<PathSelectors>;
 
-        class Path<A = PathAttributes> extends dia.Element<A> {
+        class Path extends dia.Element<PathAttributes> {
         }
 
         interface PolygonSelectors  extends dia.Cell.Selectors {
@@ -2040,7 +2040,7 @@ export namespace shapes {
 
         type PolygonAttributes = dia.Element.GenericAttributes<PolygonSelectors>;
 
-        class Polygon<A = PolygonAttributes> extends dia.Element<A> {
+        class Polygon extends dia.Element<PolygonAttributes> {
         }
 
         interface PolylineSelectors extends dia.Cell.Selectors {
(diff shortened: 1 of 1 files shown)


kumilingus:
Hey @danielpza, I will have to revert 2 of your PRs unfortunately. They both introduce a breaking change (it wasn't caught because Travis CI was disabled).

The problem is that due to the class parameters inference, it's not possible to do the following.
```ts
const rect = new shapes.standard.Rectangle({ myAttribute: 1 });
```

It will technically result in:
```ts
class Rectangle extends dia.Element<{ myAttribute: 1 }> {}
```

 Which makes a call like this invalid:
```ts
rect.set({ position: { x: 10, y: 10 }}); 
```

I was trying to find any solution to this, but didn't succeed ([StackOverflow Question](https://stackoverflow.com/questions/73556247/generic-constraints-are-ignored-when-inferring-class-parameters)).

types/joint.d.ts
```diff
@@ -1996,7 +1996,7 @@ export namespace shapes {
 
         type RectangleAttributes = dia.Element.GenericAttributes<RectangleSelectors>;
 
-        class Rectangle<A = RectangleAttributes> extends dia.Element<A> {
+        class Rectangle extends dia.Element<RectangleAttributes> {
```
danielpza, line 1999: Would something like this work?
```suggestion
         class Rectangle<A extends RectangleAttributes = RectangleAttributes> extends dia.Element<A> {
```

danielpza: Or perhaps

```suggestion
        class Rectangle<A = RectangleAttributes> extends dia.Element<A & RectangleAttributes> {
```

danielpza:
Perhaps this could work https://www.typescriptlang.org/play?#code/C4TwDgpgBAsiAq5oF4oG8oEMD8AuKAdgK4C2ARhAE4A0UZehpFlUAvgNwBQnAxgDaYAzoNggAwgOEAeeFAgAPYBAIATEXESQoqDUgB86TlGNQeAewKDglIj2BnKACmD54ASnSsjJkhGAALMxVnfAB5EgBLYBlaAGsIEDMAM1FNCAMAMlSkDzQvL14LKygScUlBAEZtQggAd1EJIUEpXUg9RwxMfCrWNy5SxuEKgDpfAKCOum62Pu5Cy2ASsqaAJmqCOobyya6oHtmAegOoAHV-ECgIkQCruUpKB2xOI6hZQUCiPhU5RWVvgANWhB-nIAG7KKDjIgAc38kKQUFqQkuBCSVEoEG+SQeJCwpiK1ls9hYYEwlEwYyow04A3KK1GfkCwQwZGmvS4QA

types/joint.d.ts
kumilingus: The `npx grunt ts` test still fails. This does not stop the class parameter inference.

kumilingus:
The `npx grunt ts` test would fail.

Plus I am not sure I want to change all the related methods because of this (`set`, `attr`, `prop`, and possible other). We don't know if it would introduce some other issues.

Especially when there is an easy workaround. Do not extend from `shapes.standard`, but define the link from scratch.

```ts
export class MyLink<T extends joint.dia.Link.Attributes> extends joint.dia.Link<T> {

    defaults() {
        // defaults() is not a function but an object hash in the `dia.Link`
        const linkDefaults = super.defaults as unknown as Partial<T>;
        return {
            ...linkDefaults,
            source: { x: 1, y: 1 },
            type: 'MyLink',
            attrs: {
                line: {
                    connection: true,
                    stroke: '#333333',
                    strokeWidth: 2,
                    strokeLinejoin: 'round',
                    targetMarker: {
                        'type': 'path',
                        'd': 'M 10 -5 0 0 10 5 z'
                    }
                },
                wrapper: {
                    connection: true,
                    strokeWidth: 10,
                    strokeLinejoin: 'round'
                }
            }
        };
    }

    preinitialize(): void {
        this.markup = [{
            tagName: 'path',
            selector: 'wrapper',
            attributes: {
                'fill': 'none',
                'cursor': 'pointer',
                'stroke': 'transparent',
                'stroke-linecap': 'round'
            }
        }, {
            tagName: 'path',
            selector: 'line',
            attributes: {
                'fill': 'none',
                'pointer-events': 'none'
            }
        }];
    }
}

interface MyAttributes extends joint.dia.Link.Attributes {
    attribute1: boolean;
}

class MyChildLink extends MyLink<MyAttributes>  {

    defaults() {
        return {
            ...super.defaults(),
            type: 'MyChildLink',
            attribute1: true
        };
    }
}
```

Note that I am ok to re-open those PRs and merge them in `v4.0`.

danielpza:
I understand, I'll see if I can find a workaround

PR #1808, opened 2022-09-14
mvc.Listener

MartinKanera:
## Description

Added Listener class, tests, types, and docs.

### DIFF
--- a/docs/src/joint/api/mvc/Listener.html
+++ b/docs/src/joint/api/mvc/Listener.html
@@ -1,4 +1,4 @@
-<p>Listener is an object that allows you to add listeners to another object's events and then remove them all when needed. Constructor accepts any number of callback arguments - invoking callback function puts provided callback arguments first with rest of the event parameters after. Context is an optional parameter that will be bound as a callback's context if provided.</p>
+<p>A listener is an object that allows you to listen for events triggered by other objects and, if needed, remove them all at once.</p>
 
 <p>Two signatures are used to listen to events:</p>
 <ol>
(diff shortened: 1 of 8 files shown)


src/mvc/index.mjs
```diff
@@ -339,3 +339,38 @@ if ($.event && !(DoubleTapEventName in $.event.special)) {
...
+        const { callbackArguments } = this;
+        // signature 1 - (object, eventHashMap, context)
+        if (args.length < 2 && typeof evt === 'object') {
+            const [context = null] = args;
+            Object.entries(evt).forEach(([eventName, cb]) => {
+                if (typeof cb !== 'function') return;
+                const functionArgs = [context];
+                // Invoke the callback with callbackArguments passed first
+                if (callbackArguments.length > 0) functionArgs.push(...callbackArguments);
+                cb = cb.bind(...functionArgs);
```
kumilingus, line 358: I would not create a new function when no `context`  nor `callbackArguments` were provided.

docs/src/joint/api/mvc/Listener.html
```diff
@@ -0,0 +1,64 @@
+<p>Listener is an object that allows you to add listeners to another object's events and then remove them all when needed. Constructor accepts any number of callback arguments - invoking callback function puts provided callback arguments first with rest of the event parameters after. Context is an optional parameter that will be bound as a callback's context if provided.</p>
```
kumilingus, line 1: > A listener is an object that allows you to listen for events triggered by other objects and, if needed, remove them all at once.

kumilingus (changes requested):

PR #1816, opened 2022-09-22
dia.LinkView, g.Line: update types

jamesgeorgewilliams:
## Description

- Add missing protected method type for `findPath` in `dia.LinkView`
- Add missing parameter for `parallel` method in `g.Line`

Both of these errors were encountered in `VSMElectronicInformationFlowView`.

Screenshot 2022-09-22 at 15 36 35
Screenshot 2022-09-22 at 15 37 30

### DIFF
--- a/types/geometry.d.ts
+++ b/types/geometry.d.ts
@@ -291,7 +291,7 @@ export namespace g {
 
         clone(): Line;
 
-        parallel(): Line;
+        parallel(distance?: number): Line;
 
         closestPoint(p: PlainPoint | string): Point;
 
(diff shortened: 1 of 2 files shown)


types/geometry.d.ts
```diff
@@ -291,7 +291,7 @@ export namespace g {
 
         clone(): Line;
 
-        parallel(): Line;
+        parallel(distance?: number): Line;
```
kumilingus, line 294: The `distance` is actually mandatory.

jamesgeorgewilliams: okay

PR #1822, opened 2022-09-30
attributes: fontSize and strokeWidth use calc()

MartinKanera:


demo/dynamic-font-size/src/index.js
```diff
@@ -0,0 +1,88 @@
...
+
+const ResizeTool = joint.elementTools.Control.extend({
+    getPosition: function(view) {
+        const model = view.model;
+        const { width, height } = model.size();
+        return { x: width, y: height };
+    },
+    setPosition: function(view, coordinates) {
+        const model = view.model;
+        model.set(
```
kumilingus, line 28: I would simplify this.
`model.position(Math.max(coordinates.x, 1), Math.max(coordinates.y, 1))`
No need to send `ui` /  `tool` flags in the demo.

kumilingus:
The new attributes have to be added here: https://github.com/clientIO/joint/blob/master/docs/src/joint/api/dia/attributes/intro.html#L149

PR #1849, opened 2022-10-20
fix isEmbeddedIn opt

kelanwu:
Should use `assign` instead of `defaults`, otherwise `cell.isEmbeddedIn(parent, { deep: false })` cannot set opt.deep to false.


kumilingus:
Thanks for contribution! The [test](https://github.com/clientIO/joint/blob/master/test/jointjs/basic.js#L1160) is also wrong 🤦. It must be `assert.notOk()`.

PR #1929, opened 2022-12-05
dia.Paper: fix magnet contextmenu handling

Geliogabalus:
## Description

Fix of the `element:magnet:contextmenu` event handling in dia.Paper. Preserve events hierarchy so it is possible to use `stopPropagation()` in `element:magnet:contextmenu` handler to prevent triggering of the `element:contextmenu` and `cell:contextmenu` events.

### DIFF
--- a/src/dia/Paper.mjs
+++ b/src/dia/Paper.mjs
@@ -2058,8 +2058,8 @@ export const Paper = View.extend({
 
         if (evt.button === 2) {
             this.contextMenuFired = true;
-            const event = $.Event(evt, { type: 'contextmenu', data: evt.data });
-            this.contextMenuTrigger(event);
+            const contextmenuEvt = $.Event(evt, { type: 'contextmenu', data: evt.data });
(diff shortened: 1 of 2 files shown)


src/dia/Paper.mjs
```diff
@@ -2319,12 +2320,21 @@ export const Paper = View.extend({
 
     onmagnet: function(evt) {
 
-        this.magnetEvent(evt, function(view, evt, _, x, y) {
-            view.onmagnet(evt, x, y);
-        });
+        if (evt.button === 2) {
+            this.contextMenuFired = true;
+            this.magnetContextMenuFired = true;
+            const event = $.Event(evt, { type: 'contextmenu', data: evt.data });
```
kumilingus, line 2326: Can we call it `contextmenuEvt`? It's hard to tell which one is which.
```js
if (event.isPropagationStopped()) {
    evt.stopPropagation();
}
```

kumilingus (changes requested):

kumilingus (approved):

PR #1942, opened 2022-12-13
fix(css): layout.css remove unprefixed user-drag property

zbynekstara:
## Description

Remove the unprefixed `user-drag` property from `joint/css/layout.css`, since it is not supported by any browsers and some tests are throwing errors.

## Motivation and Context

Note: The prefixed version was recently introduced in #1933.

```diff
.joint-element * {
  -webkit-user-drag: none;
- user-drag: none;
}
```

### DIFF
--- a/css/layout.css
+++ b/css/layout.css
@@ -58,7 +58,6 @@ the whole group of elements. Each plugin can provide its own stylesheet.
 
 .joint-element * {
    -webkit-user-drag: none;
-   user-drag: none;
 }
 
 .joint-element .scalable * {


alexandernst:
This patch is incorrect imho. The `user-drag` property should **stay** along with the vendored one! If only the vendored one is present, the property won't worn on non-chromium based browsers.

alexandernst:
Nevermind, I just saw caniuse.com, indeed, the `user-drag` shouldn't be here!

PR #1948, opened 2022-12-16
feat: monorepo architecture (part 1) [master]

zbynekstara:
## Description

Switching to a monorepo architecture. This PR is part 1 - converting the joint repository into a package (at the path `/packages/joint-core`).

The bulk of this PR deals with reconnecting all dependency paths to keep pointing to `/node_modules` (which stays in the root directory).

Additionally, three proof-of-concept additional packages are added (`joint-decorators`, `joint-elements`, `joint-element-tools`), as well as two proof-of-concept demos which combine multiple packages (`/examples` folder).

## Motivation and Context

Monorepo architecture allows users to only `require` the parts of our application which they actually need.

The current layout of JointJS already has several layers of functionality which can be used more-or-less separately. Those layers will be formalized by monorepo packages:

- Geometry
- Vectorizer (requires Geometry)
- core Joint functionality (requires Vectorizer) = `@joint/core`
- various `joint.shapes` plugins (require core Joint)
- `joint.layout.DirectedGraph` plugin (requires core Joint)

In fact, we are already emulating the monorepo functionality by having several `dist` files, with the different layers separated.

Note: The monorepo architecture is implemented via NPM Workspaces, which require at least NPM v7.0.0 (for `npm install` support). However, see their changelog for detailed information about the rollout of the Workspaces feature to other commands.


zbynekstara:
Replaced by https://github.com/clientIO/joint/pull/2062 (squashed, targeted at `dev` branch)

PR #1978, opened 2023-01-09
feat(util.breakText): support lineHeight in px

MartinKanera:
feat: return test

## Description

Allow usage of pixel units when breaking text with `util.breakText` function. Pixels are a default unit when no unit is provided.
```js
const styles = {
    fontSize: 14,
    fontFamily: 'Arial, helvetica, sans-serif'
};
const t = 'This is very very very very very very very very very very very very very very very very very very very very very very long text';
// valid
const lineHeight = 20;
// valid as well
// const lineHeight = '20px';
const cell = new joint.shapes.standard.Rectangle({
    size: { width: 50, height: 250 },
    attrs: {
        text: {
            text: t,
            textWrap: {},
            ...styles,
            lineHeight
        }
    }
});
graph.addCell(cell);
```
fixes: *util.breakText - incorrect result when custom `lineHeight` provided*

## Motivation and Context

Easier and cleaner usage of `util.breakText` function. Allow usage of pixels.

### DIFF
--- a/src/util/util.mjs
+++ b/src/util/util.mjs
@@ -524,6 +524,23 @@ function splitWordWithEOL(word, eol) {
     return eolWords.filter(word => word !== '');
 }
 
+function getLineHeight(heightValue, textElementHeight) {
+    if (heightValue === null) {
+        // Default 1em lineHeight
+        return textElementHeight;
+    } 
+
+    switch (heightValue.unit) {
+        case 'em':
+            return textElementHeight * heightValue.value;
+        case 'px':
+        case '':
+            return heightValue.value;
+        default:
+            return textElementHeight;
+    }
+}
+
 export const breakText = function(text, size, styles = {}, opt = {}) {
 
     var width = size.width;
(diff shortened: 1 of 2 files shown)


src/util/util.mjs
```diff
@@ -718,23 +735,17 @@ export const breakText = function(text, size, styles = {}, opt = {}) {
 
             if (lineHeight === undefined && textNode.data !== '') {
 
-                var heightValue;
-
                 // use the same defaults as in V.prototype.text
                 if (styles.lineHeight === 'auto') {
-                    heightValue = { value: 1.5, unit: 'em' };
+                    lineHeight = getLineHeight({ value: 1.5, unit: 'em' }, textElement.getBBox().height);
```
kumilingus, line 740: The bounding box (`textElement.getBBox().height`) is calculated when the `lineHeight` is provided in pixels but the value is never used.

This should be optimized somehow e.g.
```
function getLineHeight(heightValue, textElement) {}
```

PR #1998, opened 2023-01-20
types(mvc): Fix TS error in ModifiedCallback signature

zbynekstara:
## Description

The changed line causes the following error "TS1256: A rest element must be last in a tuple type." This is caused by the two rest operators for `...args`. I see what the intention was (a variable number of CallbackArgs followed by a variable number of arguments from EventCallback) but I don't see a way to communicate that in valid typescript.

### DIFF
--- a/types/joint.d.ts
+++ b/types/joint.d.ts
@@ -3372,7 +3372,7 @@ export namespace mvc {
         protected onRemove(): void;
     }
 
-    type ModifiedCallback<CallbackArgs extends any[], EventCallback extends Callback> = (...args: [...CallbackArgs, ...Parameters<EventCallback>]) => any;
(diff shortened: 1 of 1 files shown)


MartinKanera:
The original code should be valid for TS 4.0 and above. Just quick typescript definitions in [playground](https://www.typescriptlang.org/play?ts=4.0.5#code/C4TwDgpgBAwghgGwQIzgYwNZQLxQBQB0RcATgOYDOAXFHAHYgDaAugJQ4B8tDA3AFB9QkKAFkA9gBMAlgDMpECfCSpMAHiUp0GAILkKUCAA9gEOhP30mzADRQAogDdTwDSqxGTZ-a61dchYj0aRiICH0xdSltQgAVSOABbCBMSClVHZ3CMDjZObhAeIA).

zbynekstara:
Thank you for checking, I am closing this issue.

PR #2010, opened 2023-01-26
docs(tutorials): add portProp example to ports tutorial

jamesgeorgewilliams:
- update `Working with ports` tutorial with `portProp` example
Screenshot 2023-01-31 at 8 35 15

### DIFF
--- a/tutorials/ports.html
+++ b/tutorials/ports.html
@@ -255,7 +255,7 @@ model.addPorts([
             <h3 id="port-prop">Setting port properties</h3>
 
             <p>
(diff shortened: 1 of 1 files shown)


tutorials/ports.html
```diff
@@ -250,6 +250,42 @@ <h3 id="create">Creating elements with ports</h3>
         attrs: { label: { text: 'out' }}
     }
 ]);
+</code></pre>
+
+            <h3 id="port-prop">Setting port properties</h3>
+
+            <p>
+                If you are already familiar with the <a href="/docs/jointjs#dia.Element.prototype.attr" target="_blank"><code>attr()</code></a>
```
kumilingus, line 258: The counterpart of the `portProp()` function is the `prop()` function, not `attr()`.

PR #2032, opened 2023-02-08
fix(dia.LinkView): fix incorrect absolute offset label rotation

zbynekstara:
## Description

Labels using `keepGradient` and `absoluteOffset` have unexpected rotation applied to them.

## Motivation and Context

The issue was caused by the `g.Point.offset()` method being applied on `start` point of `tangent` line. The method modifies the point in-place.

This caused the `tangent.start` point stay in this modified state beyond the intended scope (i.e. out-of-sync with `tangent.end`) - during the time of the `angle` calculation, which happens afterward. On top of the `tangent` angle, the `angle` was thus then also incorrectly adding the angle between the modified `tangent.start` position and the original `tangent.end` position.

This PR fixes the issue by cloning the `tangent.start` point and applying the offset method to the clone. This also applies to the fallback algorithm, which uses `path.start` in absence of a tangent line.

### Screenshots:

Correct rotation while on line:

Screenshot 2023-02-07 at 17 45 37

Incorrect rotation when dragged away from line:

Screenshot 2023-02-07 at 17 45 52 | Screenshot 2023-02-07 at 17 46 01

Correct rotation when dragged away from the line thanks to this PR:

Screenshot 2023-02-10 at 11 14 58

### DIFF
--- a/test/jointjs/links.js
+++ b/test/jointjs/links.js
@@ -1200,6 +1200,67 @@ QUnit.module('links', function(hooks) {
             v1.pointerup(event);
         });
 
+        QUnit.test('label - keepGradient', function(assert) {
+
+            assert.expect(2);
+
+            var r1 = new joint.shapes.basic.Rect({ position: { x: 50, y: 50 }, size: { width: 50, height: 50 }});
+            var r2 = r1.clone().translate(250);
+
+            this.graph.addCell([r1, r2]);
+
+            var l0 = new joint.shapes.standard.Link({
+                source: { id: r1.id }, // left anchor = (100, 75)
+                target: { id: r2.id }, // right anchor = (300, 75)
+                labels: [{
+                    position: {
+                        distance: .5, // midpoint = (200, 75)
+                        args: {
+                            keepGradient: true
+                        }
+                    },
+                    attrs: { text: { text: 'test label' }} // text anchor = center = midpoint
+                }]
+            });
+
+            this.graph.addCell(l0);
+
+            var v0 = this.paper.findViewByModel(l0);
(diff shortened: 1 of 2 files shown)


test/jointjs/links.js
```diff
@@ -1200,6 +1200,67 @@ QUnit.module('links', function(hooks) {
...
+                    },
+                    attrs: { text: { text: 'test label' }} // text anchor = center = midpoint
+                }]
+            });
+
+            this.graph.addCell(l1);
+
+            var v1 = this.paper.findViewByModel(l1);
+            v1.options.interactive = { labelMove: true };
+            var event = { currentTarget: v1.$('.label')[0], type: 'mousedown' };
```
kumilingus, line 1256: `event` variable was already declared.

test/jointjs/links.js
```diff
@@ -1200,6 +1200,67 @@ QUnit.module('links', function(hooks) {
...
+            v1.dragLabelStart(event, 200, 75);
+            v1.pointermove(event, 200, 25);
+            var l1labelRotation = V.decomposeMatrix(v1.el.getElementsByClassName('labels')[0].getElementsByClassName('label')[0].getCTM()).rotation;
```
kumilingus, line 1259: Please simplify to:
```js
var l1labelRotation = V.decomposeMatrix(v1.el.querySelector('.labels .label').getCTM()).rotation;
```

kumilingus: The same for `l0labelRotation`.

kumilingus (changes requested):

kumilingus (approved):

PR #2053, opened 2023-02-16
fix(connectionPoints.Boundary): disabling automatic magnet lookup via selector option

### LINKED ISSUE #2030: connectionPoints.boundary: add option to disable the geometry magnet lookup
### Discussed in https://github.com/clientIO/joint/discussions/2028

<div type='discussions-op-text'>

<sup>Originally posted by **AnonIT03** February  7, 2023</sup>
Hello everyone, I'm facing an issue with JointJS. I have an element called UMLClass, which consists of three rectangles (Name, Methods, and Attributes). When I try to connect this element with a link, it always attaches in the middle of the element. However, all my other elements attach at the optimal position on the outside based on their position. Does anyone have any suggestions on how to fix this? Thank you in advance.</div>

kumilingus:
## Description

If the `selector` option of the `boundary` link connection point is set to `false`, the connection point is found on the boundary of the `magnet` regardless of what type of SVGElement it is. Previously, it would try to find the first non-group node descendant of the `magnet` if it the `magnet` was an `SVGGroup`, and this behavior could not be stopped.

## Motivation and Context

See comment: https://github.com/clientIO/joint/discussions/2028#discussioncomment-4901097

Fixes #2030.

### DIFF
--- a/docs/src/joint/api/connectionPoints/boundary.html
+++ b/docs/src/joint/api/connectionPoints/boundary.html
@@ -39,7 +39,7 @@
     <tr>
         <th>selector</th>
         <td><i>string</i></td>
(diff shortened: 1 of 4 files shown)


zbynekstara (approved):
Looks good to me

PR #2057, opened 2023-02-21
fix(demo.custom-embedding): separate demo, better captions, add readme

zbynekstara:
## Description

The custom embedding demo (`/demo/embedding/custom-embedding.html`) is separated into its own folder, with additional improvements to make its purpose more clear.

## Motivation and Context

The custom embedding demo has multiple issues.
- It requires special steps to be run (`grunt build:bundles` needs to be run), unlike other demos in the `/demo/embedding` folder.
  - Fixed by moving the demo into its own separate folder, and adding a README file.
- It is not clear what behavior it is trying to show.
  - Fixed by switching to more intuitive custom embedding logic.
  - Fixed by changing labels on demo elements - to give a hint of what the custom embedding logic is, and how the user can see it in action.
- It is using jQuery and Backbone.
  - Fixed by switching to a DOM method (`document.getElementById()`).
- It has a typo in a property name.
  - Fixed (`customEmebedding` -> `customEmbedding`).

## Screenshots

New look:
Screenshot 2023-02-22 at 11 00 19

Old look:
Screenshot 2023-02-21 at 18 41 58


kumilingus (changes requested):
Can we:
1. remove `Backbone` and `$` from the example
2. fix typo i.e. `customEmebedding`
3. change the logic to something that makes sense e.g. embed when nearby an element or any of its children
```js
    findParentBy: (draggedElement) => {
        const draggedBBox = draggedElement.getBBox();
        const elements = graph.getElements().filter((el) => {
            let elBBox;
            if (el.get('customEmbedding')) {
                // In case of custom embedding, we need to inflate the element's bounding box
                // that includes the element's children.
                elBBox = el.getBBox({ deep: true });
                elBBox.inflate(EMBEDDING_OFFSET);
            } else {
                elBBox = el.getBBox();
            }
            return joint.g.intersection.exists(draggedBBox, elBBox);
        });
        // Sort elements by the number of ancestors. The element with the least number of ancestors
        // comes first.
        return joint.util.sortBy(elements, (el) => -el.getAncestors().length);
    }
```

kumilingus (approved):

PR #2062, opened 2023-02-27
chore: monorepo architecture (part 1) [dev]

zbynekstara:
## Description

Switching to a monorepo architecture. This PR is part 1 - converting the joint repository into a package (at the path `/packages/joint-core`).

The bulk of this PR deals with reconnecting all dependency paths to keep pointing to `/node_modules` (which stays in the root directory).

Additionally, three proof-of-concept additional packages are added (`joint-decorators`, `joint-elements`, `joint-element-tools`), as well as two proof-of-concept demos which combine multiple packages (`/examples` folder).

## Motivation and Context

Monorepo architecture allows users to only `require` the parts of our application which they actually need.

According to [Yarn documentation](https://yarnpkg.com/features/workspaces) (which also applies to NPM workspaces, as tested by me):

>Workspaces have two important properties:

>Only the dependencies depended upon by a workspace can be accessed. Said another way, we strictly enforce your workspaces dependencies. Doing this allows us to cleanly decouple projects from one another, since you don't have to merge all their dependencies in one huge unmaintainable list. We still provide tools to manage dependencies from multiple workspaces at once, but they need to be explicitly used and offer a better integration (for example `yarn add` can make suggestions for your new dependencies based on what other workspaces use, but you can override them).

>If the package manager was to resolve a range that a workspace could satisfy, it will prefer the workspace resolution over the remote resolution if possible. This is the pillar of the monorepo approach: rather than using the remote packages from the registry, your project packages will be interconnected and will use the code stored within your repository.

What this means is that when a `package.json` file mentions a package version (e.g. within the root `package.json`, `"dependencies": {"@joint/decorators": "^0.2.1"}`), NPM first checks within the shared workspace (`joint/packages` folder, as specified by the root `package.json`) to try and find a local package whose `package.json` has such a `"name"` + a `"version"` that fulfils the criterion.

Only if that fails (i.e. there is no workspace, there is no package whose `"name"` matches `"@joint/decorators"`, or there is such a package but its `"version"` is less than or equal to 0.20), then NPM looks to the registry, where it expects to find the package at the specified version.

Note that this means that packages can only be used in the above "smart" way if they have been published. Until then, they must be referenced explicitly via the `file:` protocol (e.g. `"dependencies": {"@joint/general-shapes": "file:packages/joint-general-shapes"}`).

On top of the above, which NPM and Yarn have in common, [Yarn also offers the `workspace:` range](https://yarnpkg.com/features/workspaces#workspace-ranges-workspace), which enhances the functionality:

>While Yarn automatically picks workspace resolutions when they match, there are times where you absolutely don't want to risk using a package from the remote registry even if the versions don't match (for example if your project isn't actually meant to be published and you just want to use the workspaces to better compartment your code).

>For those use cases, Yarn now supports a new resolution protocol starting from the v2: `workspace:`. When this protocol is used Yarn will refuse to resolve to anything else than a local workspace. This range protocol has two flavors:

>When a workspace is packed into an archive (whether it's through yarn pack or one of the publish commands like yarn npm publish), we dynamically replace any `workspace:` dependency by:

>The corresponding version in the target workspace (if you use *, ^, ~, or a project-relative path)

In effect, we are forcing Yarn to only look at our workspaces (thus flagging to us if anything we expect is missing, regardless if something like it has been previously published by us to Yarn), and automatically swapping the path for the version mentioned in the local dependency when publishing. (Note that in our publishing process, we only update the package version number immediately before publishing, so we must not use precise number dependencies within the `workspace:` paths - i.e. `workspace:^` not `workspace:*`! The dependency version current at the moment of the parent publishing is applied, so we could inadvertently lock to previous version of the dependency if we deployed in wrong order. Note that this doesn't help with major releases, so we will need to be extra careful there!)

-----

The current layout of JointJS already has several layers of functionality which can be used more-or-less separately. Those layers will be formalized by monorepo packages:

- Geometry
- Vectorizer (requires Geometry)
- core Joint functionality (requires Vectorizer) = `@joint/core`
- various `joint.shapes` plugins (require core Joint)
- `joint.layout.DirectedGraph` plugin (requires core Joint)

In fact, we are already emulating the monorepo functionality by having several `dist` files, with the different layers separated.

Note: The monorepo architecture is implemented via NPM Workspaces, which require at least NPM v7.0.0 (for `npm install` support). However, see their changelog for detailed information about the rollout of the Workspaces feature to other commands.


zbynekstara:
Replaced by #2227 (squashed, targeted at new dev branch)

PR #2070, opened 2023-03-06
feat(dia.attributes): add `props` special attribute for form control properties

kumilingus:
## Description

Add `props` attribute for setting various HTML Form Control properties.

Supported properties are: `checked`, `selected`, `disabled`, `readOnly`, `contentEditable`, `value`, 'indeterminate'.

## Motivation and Context

For example, the value attribute (`<input value="test"/>`) serves only as an initial value.
When working with foreign objects, you need to be able to set the value at any time through the model (instead of accessing properties through DOM elements).

```js
// When the new value is set on the model
shape.attr('input/props/value', 'test');

const inputEl = document.createElement('input');

// We internally set it as a property
inputEl.value = 'test';

// instead of setting it as an attribute
inputEl.setAttribute('value', 'test');
```


docs/src/joint/api/dia/attributes/value.html
zbynekstara, line 1: Should we also have documentation for other property attributes added by this PR? (checked, selected, disabled, readonly, contenteditable)

PR #2074, opened 2023-03-10
feat(routers): add RightAngle router

MartinKanera:
## Description

A new orthogonal router that completely ignores vertices and generates route in compliance with spacing value. In default, it determines the most suited connection point depending on the side where the anchor is located at. Optional arguments can modify this behavior - enums for all the options have been added.

Type definitions and documentation has been updated.


src/routers/rightAngle.mjs
```diff
@@ -3,62 +3,108 @@ const Directions = {
     LEFT: 'left',
     RIGHT: 'right',
     TOP: 'top',
-    BOTTOM: 'bottom'
+    BOTTOM: 'bottom',
+    ANCHOR_SIDE: 'magnet',
```
kumilingus, line 7:
```
'ANCHOR_SIDE': 'anchor-side',
'MAGNET_SIDE': 'magnet-side',
```

PR #2104, opened 2023-03-30
dia.Paper.scaleContentToFit: adding alignment options

Geliogabalus:
Added `verticalAlign` and `horizontalAlign` options for `dia.Paper.scaleContentToFit` function.

- opt.verticalAlign – a string which specifies how the content shold be vertically aligned in the transformed paper. The options are `'top'`, `'middle'` and `'bottom'`. By default the `opt.verticalAlign` is `'top'`.

- opt.horizontalAlign – a string which specifies how the content shold be horizontally aligned in the transformed paper. The options are `'left'`, `'middle'` and `'right'`. By default the `>opt.horizontalAlign` is `'left'`.

Also `dia.Paper.scaleContentToFit` is now deprecated in favor of an alias `dia.Paper.transformToFitContent` which shows what is actually happening inside the function. For example the demo [https://www.jointjs.com/demos/paper-attributes](url) shows changes in the paper attributes and now this function directly says that it changes the paper and not the content.

### DIFF
--- a/test/jointjs/dia/Paper.js
+++ b/test/jointjs/dia/Paper.js
@@ -95,6 +95,294 @@ QUnit.module('joint.dia.Paper', function(hooks) {
         });
     });
 
+    QUnit.module('transformToFitContent', function() {
+
+        hooks.beforeEach(function() {
+            const testGraph = new joint.dia.Graph();
+            testGraph.addCells([
+                {
+                    'type': 'standard.Path',
+                    'position': {
+                        'x': 75,
+                        'y': 175
+                    },
+                    'size': {
+                        'width': 100,
+                        'height': 40
+                    },
+                    'angle': 0,
+                    'id': 'a43d5761-c4d3-474d-b765-f4f899a61480',
+                    'attrs': {
+                        'body': {
+                            'd': 'M 0 0 L calc(w) 0 calc(0.8 * w) calc(h / 2) calc(w) calc(h) 0 calc(h) Z'
+                        },
+                        'label': {
+                            'text': 'joint'
+                        }
+                    }
+                },
+                {
+                    'type': 'standard.Path',
+                    'position': {
+                        'x': 200,
+                        'y': 275
+                    },
+                    'size': {
+                        'width': 100,
+                        'height': 40
+                    },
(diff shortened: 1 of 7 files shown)


types/joint.d.ts
```diff
@@ -1505,6 +1509,8 @@ export namespace dia {
 
         scaleContentToFit(opt?: Paper.ScaleContentOptions): void;
```
kumilingus, line 1510: Mark as deprecated here too.

test/jointjs/dia/Paper.js
```diff
@@ -95,6 +95,294 @@ QUnit.module('joint.dia.Paper', function(hooks) {
         });
     });
 
+    QUnit.module('transformToFitContent', function() {
+
+        hooks.beforeEach(function() {
+            const testGraph = new joint.dia.Graph();
+            testGraph.addCells([
```
kumilingus, line 102: If I can change the `standard.Path` to `standard.Rectangle` and remove the `attrs` (`d` and `text`) and the test is still passing, please simplify the JSON. It only makes the test harder to read.

test/jointjs/dia/Paper.js
```diff
@@ -95,6 +95,294 @@ QUnit.module('joint.dia.Paper', function(hooks) {
...
+                model: testGraph,
+                async: false
+            });
+        });
+
+        QUnit.test('transformToFitContent()', function(assert) {
+
+            paper.transformToFitContent();
+
+            assert.deepEqual(paper.scale(), { sx: 1.6842105263157894, sy: 1.6842105263157894 }, 'default transform scale');
```
kumilingus, line 347: Isn't that prone to rounding error? Does every browser return the same?

kumilingus (changes requested):

test/jointjs/dia/Paper.js
Geliogabalus: It depends on the browser but it is consistent in the current test environment. I can round values but in the current implementation we will see if the browser implementation will change, i don't think it happens often.

kumilingus (approved):

Issue #1118, opened 2019-06-10
`element: magnet :pointerclick` event not triggered


wyzgo:
element:magnet:pointerclick event will not trigger when ``interactive.addLinkFromMagnet = false`` is set. Or is there any other way to get the magnet?

wyzgo:
I am not sure if it is the right way. My solution to this problem is:

kumilingus:
Hi @wyzgo, this is a nice workaround for sure. It's still a bug though. Thanks for reporting this.

kumilingus:
Fixed in #2110.

PR #2110, opened 2023-04-01
feat(dia.Paper): events handling improvements

kumilingus:
## Description

- **feat(dia.Paper)**: add `element:magnet:pointerdown`, `element:magnet:pointermove` and `element:magnet:pointerup`

  New set of events are now triggered when the user interacts with a magnet. This together with `preventDefaultInteraction()` allows anyone to implement a custom magnet interaction *e.g. instead of creating a new link, it can create a link connected to an element*.

- **feat(dia.Paper)**: better event handling of form control elements inside foreign objects

  When the form inputs are part of the element markup (inside the `<foreignObject>`) the default JointJS interaction should be prevented e.g you don't want the element to be moved while your focus is at a `<textarea>`. You need to be able to select the text inside instead.

- **feat(dia.Paper)**: add `preventDefaultViewAction` option

  When you use form controls (input, select, textarea) in your diagram, you might want to blur the active document element when the user clicks on the element view. By default, the default browser action is prevented when clicking an element view. This option allows you to run the default.

- **fix(dia.Paper)**:  event handlers now correctly receive `dia.Event` on touch screens , `originalEvent` now always points to a native event

  The `dia.Event` send to event handlers is always created from native events i.e. `originalEvent` property points to a source native event.

- **fix(dia.Paper)**:  #1118

  If a magnet is not valid, and the user just clicks on the magnet the `element:magnet:pointerclick` is fired.

- **feat(dia.LinkView)**:  fix label dragging on touch screens in the async mode

  It was not possible to drag a label on touch devices while the async mode was turned on. This was because the label gets re-rendered after each label move. Removing the original target from the DOM caused the `touchmove` events stop triggering.

- **feat(dia.CellView)**: add `preventDefaultInteraction()`  and `isDefaultInteractionPrevented()` 

  Adds API to dynamically prevent default JointJS interaction (such as element and link movement, adding a link from magnets, and label dragging). The default interaction now can be disabled based on the event (e.g. was a `shift` key pressed? what was the `target` DOM element of the event?) - as compared to the `paper.options.interactive` feature, which has access only to the cell view).
  ```js
  paper.on('element:pointerdown', (elementView, evt) => {
      if (evt.shiftKey) {
          // prevent element movement
          elementView.preventDefaultInteraction(evt);
          // start drawing selection lasso
      }
  });
  ```

- **fix(util)**: fix `normalizeEvent` to always return a `dia.Event`.

  Make sure the utility always returns a `dia.Event` (`jQuery.Event`) on touch devices. It was returning a [Touch](https://developer.mozilla.org/en-US/docs/Web/API/Touch) object enriched with `TouchEvents` properties and methods),

- **test**: make sure there is a single `qunit-fixture` container, the ability to move fixtures to the viewport and offscreen

  The fixture container for all JointJS rendering tests can be now moved from offscreen to the viewport and back. It is necessary for some of the touch events tests to have their target in the browser viewport. The `document.elementFromPoint()` that we use internally to find the element under the pointer requires the point to be in the viewport (e.g. using negative coordinates always returns `null`).

- **test**: add utilities for testing touch events

  It's possible to test the framework with simulated `TouchEvents`.
  ```js
  simulate.touchstart({ target: elRect });
  simulate.touchmove({ target: elRect, clientX: 123, clientY: 987 });
  simulate.touchend({ target: elRect });
  ```

- **feat(demo.ROI)**: add new demo with foreign objects

  image

### DIFF
--- a/demo/shapes/src/foreign-object.js
+++ b/demo/shapes/src/foreign-object.js
@@ -8,13 +8,16 @@ const paper = new joint.dia.Paper({
     model: graph,
     async: true,
     cellViewNamespace: joint.shapes,
-    guard: (evt) => ['SELECT', 'INPUT', 'BUTTON'].includes(evt.target.tagName)
+    preventDefaultBlankAction: false,
+    preventDefaultViewAction: false,
 });
 
-// Remove focus from the input when the user clicks on the paper.
-paper.on('blank:pointerdown cell:pointerdown', () => {
-    document.activeElement.blur();
-});
+// Disable the default touch action on the paper.
+// This is required in order to disable the scrolling of the page when the user starts interacting
+// with the elements on the paper using touch.
+// We set the `preventDefaultViewAction` option to `false` so clicking the element will
+// blur the form control elements (input, select, textarea)
+paper.el.style.touchAction = 'none';
 
 const Example = joint.dia.Element.define('example.ForeignObject', {
     attrs: {
@@ -27,7 +30,7 @@ const Example = joint.dia.Element.define('example.ForeignObject', {
         },
         foreignObject: {
             width: 'calc(w)',
-            height: 'calc(h)',
+            height: 'calc(h)'
         }
     }
 }, {
@@ -57,16 +60,21 @@ const Example = joint.dia.Element.define('example.ForeignObject', {
                 display: 'flex',
                 flexDirection: 'column',
                 alignItems: 'center',
-                justifyContent: 'center'
+                justifyContent: 'center',
+                // Prevents the browser from scrolling the page when the user
+                // tries to touch the foreignObject.
+                // Works in Chrome, but not in Safari. If set on the paper,
+                // it works everywhere.
+                // touchAction: 'none'
             },
             children: [{
                 tagName: 'span',
-                textContent: 'First Name'
+                textContent: 'First Name',
             }, {
                 tagName: 'input',
                 selector: 'firstname',
                 attributes: {
-                    'type': 'input',
+                    'type': 'text',
                     'name': 'firstname'
                 },
                 style: {
@@ -80,7 +88,7 @@ const Example = joint.dia.Element.define('example.ForeignObject', {
                 tagName: 'input',
                 selector: 'lastname',
                 attributes: {
-                    'type': 'input',
+                    'type': 'text',
                     'name': 'lastname'
                 },
                 style: {
@@ -90,14 +98,6 @@ const Example = joint.dia.Element.define('example.ForeignObject', {
             }]
         }]
     }]
-}, {
-    attributes: {
-        value: {
-            set: function(text, _, node) {
-                if ('value' in node) node.value = text;
-            }
-        }
-    }
 });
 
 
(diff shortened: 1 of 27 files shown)


demo/shapes/src/foreign-object.js
```diff
@@ -57,16 +60,21 @@ const Example = joint.dia.Element.define('example.ForeignObject', {
...
+                // Prevents the browser from scrolling the page when the user
+                // tries to touch the foreignObject.
+                // Works in Chrome, but not in Safari. If set on the paper,
+                // it works everywhere.
+                // touchAction: 'none'
             },
             children: [{
                 tagName: 'span',
-                textContent: 'First Name'
+                textContent: 'First Name',
```
zbynekstara, line 72: trailing commas are inconsistent in this file (line 11, 60, 64, 70, 84) - do we want to have them, not have them, or does it not matter?

docs/src/joint/api/dia/Paper/prototype/options/preventDefaultViewAction.html
zbynekstara, line 1: Shouldn't this be called `preventDefaultViewAction` since we have `paper.preventDefaultInteraction()`?

kumilingus: No. It's analogous to [preventDefaultBlankAction](https://resources.jointjs.com/docs/jointjs/v3.6/joint.html#dia.Paper.prototype.options.preventDefaultBlankAction). It prevents the browser default behavior. Not the default JointJS interaction.

demo/shapes/src/foreign-object.js
kumilingus: We don't enforce it yet, but it should be consistent (at least within the file).

src/dia/ElementView.mjs
```diff
@@ -607,24 +611,28 @@ export const ElementView = CellView.extend({
 
     pointerdown: function(evt, x, y) {
 
-        if (this.isPropagationStopped(evt)) return;
```
zbynekstara, line 610: possibly a breaking change?

kumilingus: Good question.

The propagation is stopped now in `Paper.mjs` in `pointerdown()` ([here](https://github.com/clientIO/joint/pull/2110/files/a1680f8490115aa3471afde4c3d354254157a402#diff-cad1adaf695b5bd9fb93a5a9d595e003286a15527816ce6ee7d74e014aa68cb4R2170)).

The `cellView.pointerdown()` is not called if the event propagation has been stopped on `event` or `element:magnet:pointerdown`.

*Note: it seemed to be a dead condition anyway (I wasn't able to enter it).*

src/dia/Paper.mjs
```diff
@@ -310,10 +313,6 @@ export const Paper = View.extend({
         'mouseleave .joint-cell': 'mouseleave',
         'mouseenter .joint-tools': 'mouseenter',
         'mouseleave .joint-tools': 'mouseleave',
-        'mousedown .joint-cell [event]': 'onevent', // interaction with cell with `event` attribute set
-        'touchstart .joint-cell [event]': 'onevent',
```
zbynekstara, line 314: possibly a breaking change to remove these events?

src/dia/Paper.mjs
```diff
@@ -2097,27 +2106,85 @@ export const Paper = View.extend({
...
+                    return;
+                }
+                evt.data = magnetEvt.data;
+            }
+        }
+
+        if (isContextMenu) {
             this.contextMenuFired = true;
-            const contextmenuEvt = $.Event(evt, { type: 'contextmenu', data: evt.data });
+            const contextmenuEvt = $.Event(evt.originalEvent, { type: 'contextmenu', data: evt.data });
```
zbynekstara, line 2178: breaking change?

test/jointjs/paper.js
```diff
@@ -2119,6 +2187,35 @@ QUnit.module('paper', function(hooks) {
...
+                paper.on(eventName, function(view, evt) {
+                    assert.ok(evt.originalEvent instanceof MouseEvent);
+                });
+            });
+
+            paper.options.clickThreshold = 1;
+            el.attr(['body', 'magnet'], 'passive');
+            simulate.mousedown({ el: elRect });
+            simulate.mousemove({ el: elRect });
+            simulate.mouseup({ el: elRect });
```
zbynekstara, line 2215: missing assert after these?

types/joint.d.ts
```diff
@@ -1428,6 +1433,9 @@ export namespace dia {
         $grid: JQuery;
         $background: JQuery;
 
+        GUARDED_TAG_NAMES: string[];
+        FORM_CONTROLS_TAG_NAMES: string[];
```
zbynekstara, line 1437: do we need to include these? we have a few other constants on Paper for which we don't have types

src/dia/Paper.mjs
kumilingus: It was a bug. The `originalEvent` must point to a native event or undefined.

https://api.jquery.com/category/events/event-object/
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jquery/misc.d.ts#L4702

zbynekstara:
Looks good to me

Geliogabalus (approved):

PR #2113, opened 2023-04-03
feat(routers.rightAngle): add new router

kumilingus:
## Description

A new orthogonal router that completely ignores vertices and generates routes according to the `margin` value. By default, it determines the most suitable route depending on the side where the anchor is located (or the magnet location if the magnet is a port).

The `oneSide` router has been deprecated. It could be easily replicated now with the `rightAngle` router.

Add new **flowchart** demo:
image


docs/src/joint/api/routers/rightAngle/intro.html
```diff
@@ -0,0 +1,38 @@
...
+<table>
+    <tr>
+        <th>margin</th>
+        <td><i>number</i></td>
+        <td>The minimum distance between the route and the element. The default is <code>20</code>.</td>
+    </tr>
+    <tr>
+        <th>sourceDirection</th>
+        <td><i>rightAngle.Directions</i></td>
+        <td>The direction of the link from the source anchor. The default is <code>rightAngle.Directions.AUTO</code></td>
```
jamesgeorgewilliams, line 23: I would add a period(.) at the end of this sentence, and on line:28 of this file too.

PR #2134, opened 2023-04-14
dia.Paper: autoFreeze option

Geliogabalus:
Added an option which allows paper to freeze itself after all updates are done. It prevents invoking `viewport` callback when there are no updates to the views.


alexandernst:
I’m wondering if the code could be benchmarked with and without this patch in order to figure out how much it will benefit CPU usage and battery life (I’m sure it will be something noticeable) 🤔

kumilingus:
@alexandernst I don't how to measure the energy consumption, but when this option is enabled, it will not call `viewport` function **X** number of cells **every** animation frame.

alexandernst:
Performance metrics are looking **extremely good**. Kudos for the great work!

PR #2224, opened 2023-06-19
feat(routers.RightAngle): support user defined vertices

MartinKanera:
## Description

Until now the `RightAngle` router didn't use the vertices that were provided. This change allows users to define vertices through which the link is supposed to go. The `RightAngle` router itself works the same way as it did, the only difference is that it takes in account the provided vertices and finds route between them.

## Motivation and Context

A simple upgrade of the algorithm that enables you to use the `RightAngle` router in more ways.

image

### DIFF
--- a/src/routers/rightAngle.mjs
+++ b/src/routers/rightAngle.mjs
@@ -71,13 +71,14 @@ function resolveForTopSourceSide(source, target, nextInLine) {
     const sy1 = sy0 + height;
     const smx0 = sx0 - margin;
     const smx1 = sx1 + margin;
+    const smy0 = sy0 - margin;
 
-    const { x: ax, y: ay } = anchor;
+    const { x: ax } = anchor;
     const { x0: tx, y0: ty } = target;
 
-    if (tx === ax && ty < ay) return Directions.BOTTOM;
-    if (tx < ax && ty < ay) return Directions.RIGHT;
-    if (tx > ax && ty < ay) return Directions.LEFT;
+    if (tx === ax && ty < sy0) return Directions.BOTTOM;
+    if (tx < ax && ty < smy0) return Directions.RIGHT;
+    if (tx > ax && ty < smy0) return Directions.LEFT;
     if (tx < smx0 && ty >= sy0) return Directions.TOP;
     if (tx > smx1 && ty >= sy0) return Directions.TOP;
     if (tx >= smx0 && tx <= ax && ty > sy1) {
@@ -104,17 +105,16 @@ function resolveForBottomSourceSide(source, target, nextInLine) {
     const sy1 = sy0 + height;
     const smx0 = sx0 - margin;
     const smx1 = sx1 + margin;
-    const smy0 = sy0 - margin;
     const smy1 = sy1 + margin;
 
-    const { x: ax, y: ay } = anchor;
+    const { x: ax } = anchor;
     const { x0: tx, y0: ty } = target;
 
-    if (tx === ax && ty > ay) return Directions.TOP;
-    if (tx < ax && ty > smy0) return Directions.RIGHT;
+    if (tx === ax && ty > sy1) return Directions.TOP;
+    if (tx < ax && ty > smy1) return Directions.RIGHT;
     if (tx > ax && ty > smy1) return Directions.LEFT;
-    if (tx < smx0 && ty <= sy0) return Directions.BOTTOM;
-    if (tx > smx1 && ty <= sy0) return Directions.BOTTOM;
+    if (tx < smx0 && ty <= sy1) return Directions.BOTTOM;
+    if (tx > smx1 && ty <= sy1) return Directions.BOTTOM;
     if (tx >= smx0 && tx <= ax && ty < sy0) {
         if (nextInLine.point.x < tx) {
             return Directions.RIGHT;
--- a/src/routers/rightAngle.mjs
+++ b/src/routers/rightAngle.mjs
@@ -681,7 +681,6 @@ function routeBetweenPoints(source, target) {
                 if ((y < tcy || !isSourceEl) && y > tmy0 && sox > tx1) {
                     y = tmy1;
                 }
-
                 return [
                     { x: sox, y },
                     { x: tox, y },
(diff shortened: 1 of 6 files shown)


frnora:
Hi Martin,

Here is a quick summary of the issues I found while using the new version of the router. All the screenshots are taken with shapes.standard.Rectangle with a custom Link inheriting from shapes.standard.Link.

Issues:

1. The target marker's angle is incorrect when the link is vertical. Seems like the tangent of the line segment is not calculated correctly or something like that:

frnora:
Hi Martin,

Here is a quick summary of the issues I found while using the new version of the router. All the screenshots are taken with `shapes.standard.Rectangle` with a custom Link inheriting from `shapes.standard.Link`.

Issues:

1. The target marker's angle is incorrect when the link is vertical and the target is a point. Seems like the tangent of the line segment is not calculated correctly or something like that. I assume this also affects the source marker but as my link has none, I did not test.
marker

2. The link loops when it is horizontal and the target is a point and never un-loops, even if the anchors are perfectly vertically aligned. One would expect a straight line then.
loops

3. Sometimes the route-change is not rounded but angled, despite the fact that I am using `rounded` connector. Happens around a vertex.
angled

4. The link is crossing the element. (you can once again see a not-so-round route change where the vertex is)
crossing

5. This, I assume, happens in the `orthogonal` and the original `rightAngle` one as well but I thought I would mention it. What is the reason  for these not fancy overlaps? The right side is the very same link but the vertex is moved 10 points to the left. The routing looks much better.
route

The code for point 4 (I assume it will be reproducible with it on your setup):
```js
let rect = new shapes.standard.Rectangle();
rect.isReadOnly = () => false;
rect.resize(220, 60);
rect.position(400, 400);
rect.addTo(this.#graph);

let rect2 = new shapes.standard.Rectangle();
rect2.isReadOnly = () => false;
rect2.resize(220, 60);
rect2.position(20, 200);
rect2.addTo(this.#graph);

let link = new SingleLink();
link.vertices([{
    x: 530,
    y: 320
}, {
    x: 130,
    y: 320
}]);
link.source(rect);
link.target(rect2);
link.addTo(this.#graph);
```

Thanks for the attention!

frnora:
An additional note: `view.getTangentAtRatio(0)` always returns `NaN` for `start.x` and `start.y` so something seems to be off there indeed. It works fine for every other router I tried (`manhattan`, `orthogonal`, current `rightAngle`).

kumilingus:
This usually happens when you define a path with two coincident points, e.g. `M 0 0 L 20 20 L 20 20`.

frnora:
But that's not the case. I am just moving the arrowhead around. As soon as I place it to point, the tangent calc breaks. It happens only with this version.

This is what I mean:

https://github.com/clientIO/joint/assets/129097753/3236c2e6-596f-408b-8970-c38385d154cb


Notice how plus-sign tool gets placed at 0, 0 instead of at-the-source. That's because `getTangentAtRatio` bugs out. Doesn't happen with any other router.

Actually, it gets placed at the source in one case (around 6 secs into the video) but it is in wrong position. It should be from the other side of the "arrowhead".
Screen Shot 2023-07-12 at 12 33 18

E.g. this is the current `rightAngle`. Works as expected:
https://github.com/clientIO/joint/assets/129097753/9d56a9f2-df1e-46ad-a722-c5e25df720bd

kumilingus:
What is the `d` attribute of the `[joint-selector="line"]` link node when this happens?

frnora:
E.g. in case of this one:
Screen Shot 2023-07-12 at 13 35 11

the path is: `M 300 390 L 300 390 C 300 390 300 393.3333333333333 300 400 L 300 450 C 300 456.66666666666663 296.66666666666663 459.99999999999994 290 460 L 160 460`

Complete outer html:
`<path fill="none" pointer-events="none" joint-selector="line" id="v-50" stroke="#666666" stroke-width="2" stroke-linejoin="round" d="M 300 390 L 300 390 C 300 390 300 393.3333333333333 300 400 L 300 450 C 300 456.66666666666663 296.66666666666663 459.99999999999994 290 460 L 160 460" marker-end="url(#v-2325765029)"></path>`

kumilingus:
Right, the first segment of the path is not differentiable (has no tangent defined) `M 300 390 L 300 390`.

frnora:
When you are right, you are right! Thanks for looking into this!

MartinKanera:
Hey @frnora,
Sorry for taking so long to reply, I updated the rightAngle router, could you try it out and share any possible issues?

I really appreciate any help you can provide.
Martin

frnora:
Hi @MartinKanera,

Thanks a bunch for the update! I began using it an hour ago and it looks **much** better now, so far I haven't found any problem. I will keep using it and in case of any issues, I will drop a line here.

Btw, you bamboozled me real nice with hiding it behind an option, lol. Took me several minutes to figure out that I am really not using the old/current router (while having giant question mark above my head). I assume this was done to keep it consistent with the current version? Tbh, I cannot imagine who would not want to use vertices for this router... and if not, just don't use the tool.

frnora:
Update - 2 issues found so far:

1. This is the above-mentioned issues with the connections not being rounded (there is a vertex at that point): 
Screen Shot 2023-08-02 at 12 00 49

```js
let rect = new shapes.standard.Rectangle();
rect.resize(220, 60);
rect.position(700, 700);
rect.addTo(graph);

let link = new shapes.standard.Link();
link.vertices([
    {x: 900, y: 630},
    {y: 630, x: 1040},
    {x: 1040, y: 670}
]);
link.source(rect);
link.target(rect);
link.addTo(graph);
```

2. This one is not directly related to this new version, the previous version also demonstrates it:

When the source and anchor points are aligned in a way that the link's connection point to the element is the same, it creates a kind of an L-shaped link:
Screen Shot 2023-08-02 at 10 39 29

First of all, I think it looks a bit unusual, maybe it would be better to connect it to the second closest side in this case instead of the closest side to create a loop (similarly to orthogonal).

But the "real" issue is that when this happens something is not counted correctly (cos of the "same" anchor? I don't know). On my link I have another path that's connected with the link `atConnectionRatio`. When the above happens, the path is not displayed in the correct position because the matrix calculation breaks for it:
`<path joint-selector="offsetConnector" id="v-108" d="M 0 0 0 -30" stroke="#666666" stroke-dasharray="4 4" stroke-width="2" transform="matrix(NaN,NaN,NaN,NaN,1060,825)"></path>`
Screen Shot 2023-08-02 at 10 42 02
Moving the anchor a bit upper or lower, the path appears correctly:
Screen Shot 2023-08-02 at 10 44 04

MartinKanera:
Thank you very much for your feedback. I will work on this as soon as possible!

Martin

frnora:
Great update, @MartinKanera !

It is almost there. Almost. :)

1. For new links I have a wrong tangent calc for the marker. Quite possibly happening for any link that's dragged but the arrowhead most likely covers it (and I am too lazy to check, heh). I can reproduce it easily for links coming from left/right. Top/bottom is a better but it still shows the problem when the X position of the dragged end is the same as the X of the not dragged end.
Screen Shot 2023-08-09 at 10 27 41


2. I cannot reproduce the "angled" connector anymore so that's great but some direction changes are still kind of weird. This one has 3 vertices at the points it changes direction. Padding is 20 (I guess that's the default anyway?), all the rest is default. 2 of the radii are off, like they would be squashed.
Screen_Shot_2023-08-09_at_10_26_08

I think we can live with both of these things but I thought I would let you know. If you feel like looking at one of them, pick the 1st, it is more obvious. ;)

As always, if I find something else, I will report. Thank you for your hard work!

src/routers/rightAngle.mjs
```diff
@@ -130,21 +62,261 @@ function rightAngleRouter(_vertices, opt, linkView) {
...
         case 'bottom':
-            targetOutsidePoint.y = tmy1;
+            outsidePoint.y = y0 + height + margin;
             break;
     }
 
+    return outsidePoint;
+}
+
+function routeBetweenPoints(source, target, margin) {
```
kumilingus, line 293: is `margin` parameter being used?

PR #2277, opened 2023-07-28
fix(util.breakText): NO_SPACE character present in the result

### LINKED ISSUE #2252: [Bug]: Unexpected character '0' is added to broken text by util.breakText function.
## Input  
code screenshot

## Output
resultScreenshot

I have tried to debug the problem and my solution is bellow.
solution

### Version

3.6.2

### What browsers are you seeing the problem on?

Chrome

### What operating system are you seeing the problem on?

Windows

MartinKanera:
## Description

Add the character from the previous line to the beginning of the new line only if it doesn't match the NO_SPACE constant character.

Fixes #2252.

## Motivation and Context

When breaking down text, there was a `0` character present which is internally used to separate lines, in certain edge cases (when moving the last character from the previous line to the new line).

### DIFF
--- a/src/connectors/jumpover.mjs
+++ b/src/connectors/jumpover.mjs
@@ -94,7 +94,7 @@ function updateJumpOver(paper) {
 function findLineIntersections(line, crossCheckLines) {
     return util.toArray(crossCheckLines).reduce(function(res, crossCheckLine) {
         var intersection = line.intersection(crossCheckLine);
-        if (intersection && !(line.containsPoint(crossCheckLine.start) || line.containsPoint(crossCheckLine.end))) {
+        if (intersection) {
             res.push(intersection);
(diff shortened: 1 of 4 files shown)


src/connectors/jumpover.mjs
```diff
@@ -94,7 +94,7 @@ function updateJumpOver(paper) {
 function findLineIntersections(line, crossCheckLines) {
     return util.toArray(crossCheckLines).reduce(function(res, crossCheckLine) {
         var intersection = line.intersection(crossCheckLine);
-        if (intersection) {
+        if (intersection && !(line.containsPoint(crossCheckLine.start) || line.containsPoint(crossCheckLine.end))) {
```
kumilingus, line 97: 😄

PR #2319, opened 2023-08-31
g.Rect: performance improvement when using Manhattan router with Rect shapes

coyoteecd:
## Description

Rect.containsPoint is called many times with a Point instance as argument. Avoiding the new Point() call improves loading times significantly on large papers.

## Motivation and Context

Our app uses Manhattan router with shapes and links. We've had performance issues when loading a larger paper, due to the time spent performing layout. One of the issues was that the routing logic calls `Rect.containsPoint` many, many times (thousands), and this function constructs a new Point every time it is called. But, the argument it gets is *already* a Point instance, so creating a copy just for the comparison is not necessary.

### DIFF
--- a/src/g/rect.mjs
+++ b/src/g/rect.mjs
@@ -139,7 +139,10 @@ Rect.prototype = {
 
     // @return {bool} true if point p is inside me.
     containsPoint: function(p) {
-        p = new Point(p);
+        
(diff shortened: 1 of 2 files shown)


kumilingus:
Thanks!

PR #2330, opened 2023-09-08
fix(dia.LinkView): prevent hidden doubleLinkTools to affect link bounding box

coyoteecd:
## Description

When moving around elements and links that have doubleLinkTools: true in their LinkView but link length is short enough for the secondary tool to be hidden, the tool position is not updated. 

This causes miscalculations in paper size and paddings in various scenarios:
- the secondary tool starts at 0,0 and does not get moved when the LinkView is repositioned
- shortening the link while in bottom-right corner of a large paper hides the secondary tool. Then moving the link and all other elements towards top-left does not cause the paper to autoshrink because the tool is left behind at the position where it was before being hidden.


kumilingus:
If I understand correctly, you are setting the position of the second tool to the same position as the first tool (so it doesn't affect the paper content bounding box).

I wonder if it would be better (cleaner and more efficient) to change `visibility: hidden` to `display: none`.

coyoteecd:
Setting `display: none` works equally well, I updated my change.

Issue #2297, opened 2023-08-11
[Bug]: Cache misses because of wrong assumptions in the ordering of hashmap keys


alexandernst:
### What happened?

I saw that in `src/dia/attributes/index.mjs`, in the `attributesNS` object, the `set` function of the `text` object is doing some sort of hashing / caching using `JSON.stringify` and passing a string and an object to it (`JSON.stringify([text, textAttrs]);`). The problem with this approach is that hashmap keys don't have a guaranteed order, which means that this code is most probably missing the textHash / cache most often than not.

### Version

3.7.5

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari

### What operating system are you seeing the problem on?

Mac

kumilingus:
Looking at the code, it would be relatively easy to store the text attributes one by one in a variable instead of using the `util.pick` method. I'm just trying to guess how critical this problem is.

alexandernst:
Hey! It's not critical at all. In fact, worst case scenario the cache will be missed and the text will be rerendered. I can't create a demo that will reproduce the bug. It did happened to me a couple of times (in the span of a week), but it's very random.

kumilingus:
Fixed in #2333.

PR #2333, opened 2023-09-10
fix(dia.attributes): fix ambiguous order of cache keys in text attribute

kumilingus:
Fore more details, see #2297.

### DIFF
--- a/src/dia/attributes/index.mjs
+++ b/src/dia/attributes/index.mjs
@@ -311,41 +311,54 @@ const attributesNS = {
             return !attrs.textWrap || !isPlainObject(attrs.textWrap);
         },
         set: function(text, refBBox, node, attrs) {
-            var $node = $(node);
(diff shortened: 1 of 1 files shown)


src/dia/attributes/index.mjs
```diff
@@ -311,41 +311,54 @@ const attributesNS = {
...
             if (cache === undefined || cache !== textHash) {
                 // Chrome bug:
-                // Tspans positions defined as `em` are not updated
+                // <tspan> positions defined as `em` are not updated
                 // when container `font-size` change.
                 if (fontSize) node.setAttribute('font-size', fontSize);
                 // Text Along Path Selector
-                var textPath = textAttrs.textPath;
                 if (isObject(textPath)) {
                     var pathSelector = textPath.selector;
```
zbynekstara, line 345:
```suggestion
                    const pathSelector = textPath.selector;
```

Issue #1421, opened 2021-02-05
camelCase attrs not working always


alexandernst:
I'm playing around with `joint.elementTools.Boundary` and I tried to set the stroke dasharray attribute using the camelCase syntax, but it didn't work.

```
joint.elementTools.FTABoundary = joint.elementTools.Boundary.extend({
  attributes: {
    ...
    strokeDasharray: 8,  <-- doesn't work
    "stroke-dasharray": 8  <-- it works
    ...
  },
```

alexandernst:
Another place where this occurs:

```
joint.shapes.standard.Link.define("app.Link", {
  attrs: {
    line: {
      stroke: "red",
      strokeWidth: 2, // <---- this DOES work
      targetMarker: {
        type: "path",
        d: "M -5 -5 C 2 -5 2 5 -5 5",
        fill: "none",
        strokeWidth: 2, // <---- this DOESN'T work
      }
    },
  },
});
```

alexandernst:
un-stale

kumilingus:
Fixed in `v4.0.0`.

PR #2335, opened 2023-09-11
feat(dia.CellView)!: separate alias and calc expression evaluation fro…

kumilingus:
TBD

image


kumilingus:
Currently, it allows you to switch off the support for camel-case. But if we move the alias logic to `Vectorizer.attr()` it would solve #1421.

kumilingus:
At this point, I'm not sure if it's worth optimizing. See https://jsperf.app/disifi.

alexandernst:
🤔 it’s 11% slower on iPhone 12 mini, but indeed I don’t think it will really matter that much on desktop. Maybe on batch operations where multiple attrs of multiple cells will be modified?

kumilingus:
Obsolete now. See https://github.com/clientIO/joint/pull/2459.

PR #2372, opened 2023-10-20
chore: remove lodash3 test

kumilingus:
## Description

Removing the test for `lodash` version 3.
JointJS no longer uses `lodash` directly (only Backbone uses it).
The test causes problems in Travis and it seems easier to remove the test and then fix Travis.

The PR also fixes a typo in the test runner.


jamesgeorgewilliams:
Just as an aside, I have had issues with the `lodash` tests stalling in the same place even locally occasionally in `joint/rappid` repos, so I don't think the issue is necessarily related to `travis`.

You can remove the `lodash` line from the `qunit` task too. This is where the test is run. 
https://github.com/clientIO/joint/blob/master/grunt/config/qunit.js

PR #2414, opened 2023-12-05
refactor!:  JQuery Removal

kumilingus:
## Description

This PR removes dependency on _jQuery_.

We decided to use the jQuery library as a starting point for creating our own DOM utility. We have reduced the code to less than 30 kB (unminified).
We use it only for its implementation of event namespaces and event delegation (these are remain compatible with previous versions).
We use a `WeakMap` to store DOM elements data instead of adding hidden props to them.

This new tool is for internal use only.

_Note_: _jQuery_ remains a `devDependency` as it is used in the generated documentation and a few demos.

_Note_: a reason why we didn't use the _Cash_ library (jQuery alternative) was this [issue](https://github.com/fabiospampinato/cash/issues/238).

## Breaking Changes
- drop support for _JQuery_ selectors (now only [CSS3 selectors](https://www.w3.org/TR/selectors-3/) are recognized)
   * Attribute not equal selector (!=)
   * Positional selectors (:first; :eq(n); :odd; etc.)
   * Type selectors (:input; :checkbox; :button; etc.)
   * State-based selectors (:animated; :visible; :hidden; etc.)
   * :has(selector) in browsers without native support
   * :not(complex selector) in IE
   * custom selectors via jQuery extensions
   * Reliable functionality on XML fragments
   * Matching against non-elements
   * Reliable sorting of disconnected nodes
   * querySelectorAll bug fixes (e.g., unreliable :focus on WebKit) 
   * Selectors can not start with combinators (e.g "> div" needs to be converted to ":scope > div")
- `view.$el` and `view.$()` are no longer public properties (use `view.el` and `view.el.querySelectorAll()` instead)
- remove the following paper's protected properties: `paper.$document`, `paper.$grid` and `paper.$background`
- `utils.sortElement` returns a plain array of Elements (not `$` object)
- it's no longer possible to access `CellView` from a DOM element (`$.data(document.querySelector('.joint-cell').view`)
- `cellView.prototype.findBySelector()` removed in favor of `findNode()`, `findPortNode()` and `findLabelNode()` (resp. `findNodes()`, `findPortNodes()` and `findLabelNodes()` to find nodes referenced by a group selector).

### Keep eye on
- jQuery automatically adds units for CSS style values. Is there an API that used `$.css()` originally and does not expect the values to have units? e.g. `{ width: 100 }` (Note that SVG attributes don't require units, so this only applies to CSS styles.)


packages/joint-core/docs/src/joint/api/dia/CellView/prototype/findBySelector.html
```diff
@@ -1,12 +1,9 @@
...
+<pre class="docs-method-signature"><code>cellView.findBySelector(selector)</code></pre>
 
-<p>This method mimics jQuery element search methods in that it always returns an array. If you are only expecting to find a single subelement (most common scenario), you can get it by accessing the first element in the returned array – e.g. <code>this.findBySelector('body')[0]</code>.</p>
+<p>The method returns an array of DOM nodes matching the <code>selector</code>.</p>
 
-<p>If you do not specify a <code>selector</code>, the <code>root</code> of the CellView is returned as the only element of the return array. For backwards compatibility, the same applies also if you provide the special value <code>selector = '.'</code>. By default <code>root = this.el</code> (the wrapping <code>&lt;g&gt;</code> SVGElement of the CellView), but if you provide a different <code>root</code> (see below), then that value is returned instead.</p>
+<p>The method searches for matching sub-elements among the descendants of this <code>this.el</code>. If no such sub-elements are found, an empty array is returned.</p>
```
jamesgeorgewilliams, line 5: `this <code>this.el</code>.` - Not sure if `this` before the code block is there by accident.

packages/joint-core/demo/rough/yarn.lock
jamesgeorgewilliams, line 1: `elk`, `dgl`, `ts-demo`, etc all have a `yarn.lock`, not sure if you should keep this just to be consistent, or remove all `yarn.lock` files. I added `package-lock.json` to the `.gitignore` of those demos too in case someone uses `npm` to install deps. Maybe we could add `yarn.lock` to `.gitignore` too.


and same Q for `vuejs` demo.

packages/joint-core/demo/embedding/nested-clone.js
```diff
@@ -1,8 +1,12 @@
...
+var infoEl = document.createElement('pre');
+infoEl.style.position = 'fixed';
+infoEl.style.top = '50px';
+infoEl.style.right = '100px';
+document.body.appendChild(infoEl);
 resetInfo();
 
 function resetInfo() {
-    $info.text('Hover over cells to see\nhow cloning and graph search works\non nested graphs.');
+    infoEl.innerHTML = 'Hover over cells to see\nhow cloning and graph search works\non nested graphs.';
```
zbynekstara, line 9:
```suggestion
    infoEl.textContent = 'Hover over cells to see\nhow cloning and graph search works\non nested graphs.';
```

packages/joint-core/demo/embedding/nested-clone.js
```diff
@@ -86,7 +90,7 @@ paper.on('cell:mouseenter', function(cellView) {
     i[keyGetConnectedLinks] = joint.util.toArray(graph.getConnectedLinks(cell, { deep: true })).map(function(c) {
         return c.get('name');
     }).join(',');
-    $info.text(JSON.stringify(i, '\t', 4));
+    infoEl.innerHTML = JSON.stringify(i, '\t', 4);
```
zbynekstara, line 93:
```suggestion
    infoEl.textContent = JSON.stringify(i, '\t', 4);
```

packages/joint-core/demo/graph.js
```diff
@@ -1,19 +1,18 @@
...
+infoEl.style.right = '100px';
+document.body.appendChild(infoEl);
 
 function pluck(array, attribute) {
     return array.map(function(item) { return item[attribute]; });
 }
 
 function clear() {
-    $info.text('');
+    infoEl.innerHTML = '';
```
zbynekstara, line 15:
```suggestion
    infoEl.textContent = '';
```

packages/joint-core/demo/graph.js
```diff
@@ -52,7 +51,7 @@ function info(view) {
             bfsDistance: bfsDistance
         };
 
-        $info.text(JSON.stringify(i, '\t', 4));
+        infoEl.innerHTML = JSON.stringify(i, '\t', 4);
```
zbynekstara, line 54:
```suggestion
        infoEl.textContent = JSON.stringify(i, '\t', 4);
```

packages/joint-core/demo/ports/port-layouts-comp.js
```diff
@@ -67,28 +67,57 @@ paper6.on('element:pointerclick', function(cellView) {
...
+};
+document.body.appendChild(button4);
+
+var b3 = document.createElement('b');
+b3.textContent = ' adjust height ';
+document.body.appendChild(b3);
+
+var div = document.createElement('div');
+div.innerHTML = '&nbsp;';
+
```
zbynekstara, line 122:
```suggestion
document.body.appendChild(div);
```

packages/joint-core/demo/routing/src/routing.js
```diff
@@ -105,4 +105,4 @@ $('.router-switch').on('click', function(evt) {
     }
 
     link.set('connector', { name: connector });
-});
+}, false);
```
zbynekstara, line 108: This is not necessary, `useCapture` is `false` by default

```suggestion
});
```

packages/joint-core/demo/rough/yarn.lock
zbynekstara: The issue is that not having the `yarn.lock` files in the demos produces an error in yarn:
```
Usage Error: The nearest package directory (/Users/zbynek/code/joint/packages/joint-core/demo/vuejs) doesn't seem to be part of the project declared in /Users/zbynek/code/joint.

- If /Users/zbynek/code/joint isn't intended to be a project, remove any yarn.lock and/or package.json file there.
- If /Users/zbynek/code/joint is intended to be a project, it might be that you forgot to list packages/joint-core/demo/vuejs in its workspace configuration.
- Finally, if /Users/zbynek/code/joint is fine and you intend packages/joint-core/demo/vuejs to be treated as a completely separate project (not even a workspace), create an empty yarn.lock file in it.
```

i.e. the yarn.lock contents don't have to be there, but the yarn.lock files themselves (albeit empty) should be. So a solution would be to keep all yarn.lock files but empty their contents.

packages/joint-core/docs/demo/layout/DirectedGraph/js/index.js
```diff
@@ -10,12 +10,13 @@
         model: graph
     });
 
-    $('#btn-layout').on('click', layout);
+    document.getElementById('btn-layout').addEventListener('click', layout, false);
```
zbynekstara, line 13: This should not be necessary, it is `false` by default

```suggestion
    document.getElementById('btn-layout').addEventListener('click', layout);
```

packages/joint-core/docs/demo/layout/Port/js/portRotationComp.js
```diff
@@ -56,28 +60,59 @@ paper6.on('cell:pointerclick', function(cellView, e) {
...
+b1.textContent = 'Click on Element to toggle port rotation compensation';
+document.body.appendChild(b1);
+
+var br1 = document.createElement('br');
+document.body.appendChild(br1);
 
-$('<button/>').text('-').appendTo('body').on('click', function() {
+var button1 = document.createElement('button');
+button1.textContent = '-';
+button1.onclick = function() {
```
zbynekstara, line 72: all `onclick()` calls in this file should be changed to `addEventListener()` for consistency

packages/joint-core/docs/demo/layout/Port/portRotationComp.html
```diff
@@ -20,34 +20,23 @@
...
-<script src="../../../js/lib/lodash.min.js"></script>
-
 <script src="../../../js/lib/joint.min.js"></script>
 
 <script>
     /**
      * HELPERS
      */
-    function createPaper() {
+     function createPaper() {
```
zbynekstara, line 29:
```suggestion
    function createPaper() {
```

packages/joint-core/docs/demo/layout/PortLabel/portLabel.html
```diff
@@ -20,34 +20,23 @@
...
-<script src="../../../js/lib/lodash.min.js"></script>
-
 <script src="../../../js/lib/joint.min.js"></script>
 
 <script>
     /**
      * HELPERS
      */
-    function createPaper() {
+     function createPaper() {
```
zbynekstara, line 29:
```suggestion
    function createPaper() {
```

packages/joint-core/docs/src/joint/api/dia/CellView/prototype/findBySelector.html
```diff
@@ -1,12 +1,9 @@
-<pre class="docs-method-signature"><code>cellView.findBySelector([selector, root, selectors])</code></pre>
-<p>Return an array of subelements of this CellView which are identified by <code>selector</code>. If no such subelements are found, return an empty array.</p>
+<pre class="docs-method-signature"><code>cellView.findBySelector(selector)</code></pre>
```
zbynekstara, line 1: We are changing the method signature here but not in the code - shouldn't we acknowledge the `root` and `selectors` options here but mark them as deprecated?

packages/joint-core/docs/src/joint/api/linkTools/Button.html
```diff
@@ -26,7 +26,7 @@
         <td><i>function</i></td>
         <td>What should happen when the user clicks the button? Default is <code>undefined</code> (no interaction).<br><br>
 
-            The callback function is expected to have the signature <code>function(evt, linkView, buttonView)</code> where <code>evt</code> is a <code>jQuery.Event</code> <a href="https://api.jquery.com/category/events/event-object/">object</a>. The related link view is available inside the function as <code>this</code>. The link model is available as <code>this.model</code>.</td>
+            The callback function is expected to have the signature <code>function(evt, linkView, buttonView)</code> where <code>evt</code> is a <code>mvc.Event</code> <a href="#mvc.Event">object</a>. The related link view is available inside the function as <code>this</code>. The link model is available as <code>this.model</code>.</td>
```
zbynekstara, line 29: We don't have mvc.Event in the docs, so this link will not work

packages/joint-core/docs/src/joint/api/linkTools/Remove.html
```diff
@@ -30,7 +30,7 @@
                 linkView.model.remove({ ui: true, tool: toolView.cid });
             }</code></pre>
 
-            The callback function is expected to have the signature <code>function(evt)</code> where <code>evt</code> is a <code>jQuery.Event</code> <a href="https://api.jquery.com/category/events/event-object/">object</a>. The button view is available inside the function as <code>this</code>; the button model is available as <code>this.model</code>.</p>
+            The callback function is expected to have the signature <code>function(evt)</code> where <code>evt</code> is a <code>mvc.Event</code> <a href="#mvc.Event">object</a>. The button view is available inside the function as <code>this</code>; the button model is available as <code>this.model</code>.</p>
```
zbynekstara, line 33: We don't have mvc.Event in the docs, so this link will not work

packages/joint-core/src/util/util.mjs
```diff
@@ -1086,10 +1076,9 @@ export const getElementBBox = function(el) {
 // See http://james.padolsey.com/javascript/sorting-elements-with-jquery/.
 export const sortElements = function(elements, comparator) {
 
-    var $elements = $(elements);
-    var placements = $elements.map(function() {
+    elements = $(elements).toArray();
+    var placements = elements.map(function(sortElement) {
 
-        var sortElement = this;
```
zbynekstara, line 1092: We are relying on `this` being `sortElement` later ([here](https://github.com/clientIO/joint/pull/2414/files#diff-065c48be1d93599716d40fab9433aec775a3e3233769985f739384665dfbc0caR1090) and [here](https://github.com/clientIO/joint/pull/2414/files#diff-065c48be1d93599716d40fab9433aec775a3e3233769985f739384665dfbc0caR1095), but the new code doesn't reassign `this` as expected, right?

packages/joint-core/src/util/utilHelpers.mjs
```diff
@@ -301,7 +301,7 @@ const copyObject = (source, props, object) => {
     return object;
 };
 
-const isArrayLike = (value) => {
+export const isArrayLike = (value) => {
```
zbynekstara, line 304: Where is this used? I don't see any usages outside this file in either joint or joint plus

packages/joint-core/docs/demo/layout/DirectedGraph/js/index.js
kumilingus: it's ok to be explicit.

packages/joint-core/package.json
```diff
@@ -104,6 +101,7 @@
     "grunt-webpack": "6.0.0",
     "handlebars": "4.7.7",
     "jit-grunt": "0.10.0",
+    "jquery": "~3.7.1",
```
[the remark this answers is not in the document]
kumilingus: Because of the generated documentation.

packages/joint-core/src/mvc/Dom/Dom.mjs
```diff
@@ -0,0 +1,574 @@
+
+/*!
+ * jQuery JavaScript Library v4.0.0-pre+c98597ea.dirty
```
zbynekstara, line 3: we should not mention jQuery

packages/joint-core/tutorials/elements.html
```diff
@@ -234,7 +222,7 @@ <h3 id="example">Example</h3>
 
             <div class="paper" id="paper-elements"></div>
 
-            <pre class="line-numbers" data-line="36-48,50-68,75-92,94-108,115-129,131-150" style="padding-left:3.5em"><code>&lt;!DOCTYPE html&gt;
+            <pre class="line-numbers" data-line="32-45,47-65,72-89,91-105,112-126,128-147" style="padding-left:3.5em"><code>&lt;!DOCTYPE html&gt;
```
zbynekstara, line 225:
```suggestion
            <pre class="line-numbers" data-line="33-45,47-65,72-89,91-105,112-126,128-147" style="padding-left:3.5em"><code>&lt;!DOCTYPE html&gt;
```

packages/joint-core/tutorials/installation.html
```diff
@@ -39,30 +30,28 @@ <h1>Installation</h1>
             <p>In this section, we will learn how to install JointJS and include it on our page.</p>
 
             <p>
-                For your JointJS application to run, the JointJS library and its dependencies have to be included in the
+                For your JointJS application to run, the JointJS library have to be included in the
```
zbynekstara, line 33:
```suggestion
                For your JointJS application to run, the JointJS library has to be included in the
```

packages/joint-core/tutorials/introduction.html
```diff
@@ -61,7 +49,7 @@ <h1>Tutorials</h1>
                 additional shapes (BPMN diagrams).</p>
 
             <p>
-                Let's learn how create great diagrams with JointJS. We will start by examining a simple 
+                Let's learn how create great diagrams with JointJS. We will start by examining a simple
```
zbynekstara, line 52:
```suggestion
                Let's learn how to create great diagrams with JointJS. We will start by examining a simple
```

packages/joint-core/test/utils.js
```diff
@@ -88,14 +88,135 @@
...
+$.fn.index = function() {
+    const [el] = this;
+    return Array.prototype.indexOf.call(el.parentNode.children, el);
+};
+
+$.fn.offset = function() {
+    const [el] = this;
+    const box = el.getBoundingClientRect();
+    return {
+        top: box.top + window.pageYOffset - document.documentElement.clientTop,
```
zbynekstara, line 160: `pageYOffset` and `pageXOffset` are deprecated according to VS Code, suggests changing to `scrollY` and `scrollX`

packages/joint-core/test/utils.js
```diff
@@ -88,14 +88,135 @@
...
+
+$.fn.children = function(selector) {
+    const [el] = this;
+    if (selector) {
+        return $(Array.from(el.children).filter(child => child.matches(selector)));
+    }
+    return $(el.children);
+};
+
+$.fn.parent = function(i) {
```
zbynekstara, line 173:
```suggestion
$.fn.parent = function() {
```

packages/joint-core/src/mvc/Dom/methods.mjs
```diff
@@ -0,0 +1,211 @@
...
+    if (typeof name === 'string') {
+        if (value === undefined) {
+            const [el] = this;
+            if (!el) return null;
+            return el.getAttribute(name);
+        } else {
+            attributes = { [name]: value };
+        }
+    } else if (!name) {
+        throw new Error('no styles provided');
```
zbynekstara, line 135:
```suggestion
        throw new Error('no attributes provided');
```

packages/joint-core/tutorials/advanced.html
zbynekstara, line 1: Since we are removing the archive tutorials, the whole last paragraph of this page should also be deleted: https://github.com/clientIO/joint/pull/2414/files#diff-577ff13a7f10cd2bf942780a0b2f8e6d1e299d14886904eb9afa6bffeb5a0500L41-L47

packages/joint-core/docs/src/joint/api/dia/CellView/prototype/findNode.html
zbynekstara, line 1: Are we going to document `findPortNode` and `findLabelNode`?

zbynekstara, line 1: Now that I see the changes, I actually think we should call this fuction `findNodes`, plural, since we are returning an array, not a single element, which is different from `findPortNode` and `findLabelNode`. If/when we decide to unify the behavior of the three functions, we would be able to do it by adding a new method rather than by redefining this method via a breaking change.

packages/joint-core/docs/src/joint/api/dia/LinkView/prototype/findLabelNodes.html
```diff
@@ -0,0 +1,5 @@
+<pre class="docs-method-signature"><code>linkView.findLabelNodes(index, groupSelector)</code></pre>
+
+<p>Return an array of <i>SVGElement|HTMLElement</i> referenced by <b>groupSelector</b> of the label at given <b>index</b>. If there is no label at <b>index</b> or no DOM node was found by the <b>groupSelector</b>, <code>null</code> is returned.</p>
```
zbynekstara, line 3:
```suggestion
<p>Return an array of <i>SVGElement|HTMLElement</i> referenced by <b>groupSelector</b> of the label at given <b>index</b>. If there is no label at <b>index</b> or no DOM node was found by the <b>groupSelector</b>, an empty array is returned.</p>
```

packages/joint-core/docs/src/joint/api/dia/ElementView/prototype/findPortNodes.html
```diff
@@ -0,0 +1,5 @@
+<pre class="docs-method-signature"><code>elementView.findPortNodes(portId, groupSelector)</code></pre>
+
+<p>Return an array of <i>SVGElement|HTMLElement</i> referenced by <b>groupSelector</b> of the port identified by <b>portId</b>. If there is no port with<b>portId</b> or no <i>SVGElement|HTMLElement</i> was found by the <b>groupSelector</b>, <code>null</code> is returned.</p>
```
zbynekstara, line 3:
```suggestion
<p>Return an array of <i>SVGElement|HTMLElement</i> referenced by <b>groupSelector</b> of the port identified by <b>portId</b>. If there is no port with<b>portId</b> or no <i>SVGElement|HTMLElement</i> was found by the <b>groupSelector</b>, an empty array is returned.</p>
```

zbynekstara (approved):

PR #2425, opened 2023-12-11
refactor(shapes)!: remove devs shapes; update standard shapes

kumilingus:
## Description

_This is the second PR that removes shapes from joint-core (See https://github.com/clientIO/joint/pull/2422)._

Completely removing legacy shapes from the joint-core package.
- `devs`  were removed from `joint-core` and are implemented only as custom shapes within the demos.

### Migration guide
You can copy the shape definition from `v3.7` directly to your application.
Here are examples: [devs](https://github.com/clientIO/joint/tree/dev/demo/devs)

---

The PR also updates the `standard` shapes not to use legacy attributes such as `refWidth`, `refHeight`, `refX`, `refY`, etc.
and use native SVG attributes with `calc` expressions.

For instance `refWidth: 100%;` is replaced with `width: 'calc(w)'`.

If you change these values in your application (e.g. by changing the `refD` attribute of the path or changing the position of the label using `refY`), this is another breaking change.

### Migration guide
a. You can copy the shape definition from `v3.7` directly to your application and override the built-in ones.
b. You need to switch from `ref*` attributes to native attributes in your code and you might also need to change the attributes in any JSON you try to import (should the JSON contain modified `standard` shapes `ref*` attributes).


packages/joint-core/src/shapes/standard.mjs
```diff
@@ -98,16 +98,16 @@ export const Ellipse = Element.define('standard.Ellipse', {
 export const Path = Element.define('standard.Path', {
     attrs: {
         body: {
-            refD: 'M 0 0 L 10 0 10 10 0 10 Z',
+            d: 'M 0 0 H calc(w) V calc(h) H 0 Z',
```
zbynekstara, line 101: Shouldn't we be using normalized path data? i.e. `L` with four calc() expressions

zbynekstara: We are not doing normalizations on the data later, so the solution with V and H (and two calc expressions) is better than an alternative solution with L (and four calc expressions)

PR #2441, opened 2023-12-15
fix(examples/isometric): fixed isometric example

Geliogabalus:
Fixed the invalid attribute name


kumilingus:
I think it's better to fix it by setting `joint.config.useCSSSelectors = false`

PR #2446, opened 2023-12-19
refactor(dia.LinkView)!: remove support for legacy link view

kumilingus:
## Description

Remove support for legacy features of `joint.dia.LinkView` (It has long been deprecated due to poor performance caused by built-in link tools that had to be rendered for each link at start-up (even though they are only visible when the user hovers over them).
 - `joint.dia.Link` has no markup defined and effectively becomes an abstract class
 - drop support for semantic string markup (`connection`, `connection-wrap`, `marker-source`, `marker-target`, `marker-vertices`, `marker-arrowheads`, `link-tools` in the markup has no longer effect on the user interactivity).
 - drop support for `vertexAdd`, `vertexRemove`, `vertexMove`, `arrowheadMove`,  `useLinkTools` interactivity paper options
 - the following protected methods has been removed for LinkView: `dragConnectionStart`, `dragConnection`, `dragConnectionEnd`, `dragVertexStart`, `dragVertex`, `dragVertexEnd`
 - the paper `defaultLink` is now `standard.Link`

The `linkTools.Vertices` tools now accept 2 new options: `vertexAdding` and `vertextRemoving`

### Migration guide

Switch to `joint.shapes.standard.Link` and add link tools dynamically.

For the interactive options:
- `vertexAdd`, `vertexRemove` & `vertexMove` - add `linkTools.Vertices` to enable the user to interact with vertices

    ```js
    new linkTools.Vertices({ vertexAdding: boolean, vertextMoving: boolean, vertextRemoving: boolean })
    ```
- `arrowheadMove` - add `linkTools.Arrowhead` to enable arrowhead move
- `useLinkTools` - add `linkTools.Button` to the link view


packages/joint-core/docs/src/joint/api/dia/Link/intro.html
```diff
@@ -212,18 +176,18 @@ <h4 id="dia.Link.intro.customProperties">Custom Properties</h4>
     interactive: function(cellView) {
         if (cellView.model.get('customLinkInteractions')) {
             // only links with `customLinkInteractions: true`
-            return { vertexAdd: false };
+            return { labelMove: false };
```
zbynekstara, line 179:
```suggestion
            return true;
```

packages/joint-core/docs/src/joint/api/dia/Link/intro.html
```diff
@@ -212,18 +176,18 @@ <h4 id="dia.Link.intro.customProperties">Custom Properties</h4>
...
         }
         return true; // otherwise
```
zbynekstara, line 181:
```suggestion
        return { labelMove: false }; // otherwise
```

packages/joint-core/docs/src/joint/api/dia/Link/intro.html
```diff
@@ -212,18 +176,18 @@ <h4 id="dia.Link.intro.customProperties">Custom Properties</h4>
...
     }
 });
 
-var link1 = new joint.dia.Link({
+var link1 = new joint.shapes.standard.Link({
     //...
     customLinkInteractions: true // right-click adds a label
```
zbynekstara, line 187:
```suggestion
    customLinkInteractions: true // label can be moved
```

Issue #2107, opened 2023-03-31
[Docs]: parent() doesn't embed


alexandernst:
### What happened?

I found out that calling `parent(id)` doesn't actually embed the cell. When I running the following code:

```js
const graph = new joint.dia.Graph();
var rect = new joint.shapes.standard.Rectangle();
rect.addTo(graph);
var rect2 = rect.clone();
rect2.addTo(graph);

rect2.parent(rect.id);
console.log(rect.getEmbeddedCells());
```

I get `[]`.

If I change the code to:

```js
rect.embed(rect2);
console.log(rect.getEmbeddedCells());
```

I get `[child]`.

[The documentation](https://resources.jointjs.com/docs/jointjs/v3.6/joint.html#dia.Cell.prototype.parent) states that: 

> cell.parent(parentId [, opt])
> Set the parent property of the cell to parentId provided.

At this point I'm not sure if this is a documentation problem or a functionality problem.

As a related problem, the `opt` argument is not documented.

Working demo: https://codesandbox.io/s/jointjs-parent-embeds-bug-48cnet?file=/src/index.js

### Version

3.6.5

### What browsers are you seeing the problem on?

Safari

### What operating system are you seeing the problem on?

Mac

PR #2501, opened 2024-01-23
fix(dia.Cell): remove setter from prototype in docs and types

MartinKanera:
## Description

Remove the `Cell.parent(id)` setter from type definitions and documentation.

## Motivation and Context

`Cell.parent(id)` method is not intended to be used as a way to embed cells. To eliminate the confusion it was decided to remove it from type definitions and documentation.

Related to #2107.

### Migration guide
In case you were using the `Cell.parent(cell)` as a solution to embed cells, use the `Cell.embed(cell)` instead.

#### Example usage:
```js
const parent = joint.shapes.standard.Rectangle();
const child = joint.shapes.standard.Rectangle();
parent.embed(child);
```


alexandernst:
Isn't this missing https://github.com/clientIO/joint/blob/945b64b825e62d87b637f235df562aa841297831/src/dia/Cell.mjs#L317 ? I think that line should also be removed 🤔

MartinKanera:
We still use Cell.parent setter internally, this PR only resolves the confusion with using it as a way to embed cells.

Issue #2061, opened 2023-02-25
[Bug/Feature]: updateViewsAsync keeps being invoked for no reason (?)


alexandernst:
### What happened?

I noticed that if I create a Paper with `async: false` option and a dummy `viewport` function, the function will get called just a couple of times right after loading a graph. Further invocations of `viewport` will happen only when I interact in some way with the diagram (when I click on something, when I drag something, etc...).

However, if I create a Paper with `async: true` option, the `viewport` function won't stop being called, even if I'm not interacting at all with the diagram.

I searched for anything that might be related to `viewport` here as well as in the old google groups, but I couldn't find anything, so I'm not really sure if this is an expected behaviour or not.

So I did what I usually do: debug the code and find out what's actually going on. It turns out `updateViewsAsync` (Paper.mjs) is being invoked whenever JS has access to a new tick (`nextFrame`). This function calls `checkViewport`, which then calls `checkMountedView`, which then runs `viewportFn.call`, which is my `viewport` function.

I have a couple of questions:

1) Does it make sense for `updateViewsAsync` to have some sort of check / sleep / pause that will invoke it only when it's actually required, as in, when there are actual changes that might require to trigger a re-render. Is that even feasible? This code looks quite expensive, running it on each tick might be affecting negatively the performance (and battery life / power consumption?)

2) Is it possible to, somehow, avoid invoking the `viewport` callback if nothing changed?

### Version

3.6.x

### What browsers are you seeing the problem on?

Chrome, Safari

### What operating system are you seeing the problem on?

Mac

slavede:
This also happens for me, even though I have called 

paper.remove() at the end

It causes big performance issues if you are keep navigating to that paper. You just can't destroy it/stop it

kumilingus:
@alexandernst 

It's not a bug. It is by design.

There is a plan to extend it so the paper will put itself to sleep when idle or do some chores e.g. cleaning the paper z-index pivots. And then wake up when there are updates scheduled.

> Is it possible to, somehow, avoid invoking the viewport callback if nothing changed?

Not sure, I need to investigate. Please bear with me (I'll probably look into it after the release).

---

@slavede 

I can see there is [paper.freeze()](https://github.com/clientIO/joint/blob/master/src/dia/Paper.mjs#L1194) called when the paper is removed.

Are you sure this is happening? Could you create a demo showing the issue?

kumilingus:
Hi @alexandernst, this might work for the time being.

```js
[...]

PR #2561, opened 2024-02-29
fix(dia.Paper): don't leak animation frame request

bendemboski:
## Description

This ensures that when `Paper.resetViews()` is called, any pending animation frame request is not leaked.

## Motivation and Context

When `resetViews()` calls `_resetUpdates()` it's possible that there's an animation frame request, but `_resetUpdates()` was not checking for this and cancelling any such request, potentially causing `updateViewsAsync()` to be called when it shouldn't be.

This can cause unexpected behavior if the paper was frozen, or cause an error to be thrown if the paper is removed (since `updateViewsAsync()` with throw in such a case).

### DIFF
--- a/packages/joint-core/src/dia/Paper.mjs
+++ b/packages/joint-core/src/dia/Paper.mjs
@@ -417,6 +417,8 @@ export const Paper = View.extend({
     },
 
     _resetUpdates: function() {
+        if (this._updates && this._updates.id) cancelFrame(this._updates.id);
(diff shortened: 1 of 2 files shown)


kumilingus (approved):

kumilingus:
Perhaps a proper fix for https://github.com/clientIO/joint/issues/2061 as well.
Thank you @bendemboski.

PR #2585, opened 2024-03-22
types: fix untyped model when using ElementView/LinkView

michaelspiss:
## Description
Instead of passing a generic `Element`/`Link` type to `CellViewGeneric` this provides an optional generic to pass the model type. It would be used like the following:

```typescript
interface CustomElementAttributes extends dia.Element.Attributes {
  custom_attribute: string,
}

class CustomElement extends dia.Element<CustomElementAttributes> {
  // ...
}

class CustomElementView extends dia.ElementView<CustomElement> {
  // ...
}
```

Default behaviour is the current behaviour. The change is fully backwards compatible.

## Motivation and Context
- this.model.get(...) is checked against attributes of the model
  - return value is typed instead of `any`
  - language server knows about possible keys
- No longer throws errors ("... implicitely has any type") when used with typescript
  - Removes manual overhead of hard coding types back in after using `get()`

### DIFF
--- a/packages/joint-core/types/joint.d.ts
+++ b/packages/joint-core/types/joint.d.ts
@@ -919,7 +919,7 @@ export namespace dia {
 
         setInteractivity(value: boolean | ElementView.InteractivityOptions): void;
 
-        getDelegatedView(): ElementView<E> | null;
+        getDelegatedView(): ElementView | null;
 
         findPortNode(portId: string | number): SVGElement | null;
         findPortNode(portId: string | number, selector: string): E | null;
--- a/packages/joint-core/types/joint.d.ts
+++ b/packages/joint-core/types/joint.d.ts
@@ -913,18 +913,18 @@ export namespace dia {
         }
     }
 
-    class ElementView extends CellViewGeneric<Element> {
+    class ElementView<E extends Element = Element> extends CellViewGeneric<E> {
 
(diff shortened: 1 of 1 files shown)


packages/joint-core/types/joint.d.ts
```diff
@@ -913,18 +913,18 @@ export namespace dia {
...
-    class ElementView extends CellViewGeneric<Element> {
+    class ElementView<E extends Element = Element> extends CellViewGeneric<E> {
 
-        update(element?: Element, renderingOnlyAttrs?: { [key: string]: any }): void;
+        update(element?: E, renderingOnlyAttrs?: { [key: string]: any }): void;
 
         setInteractivity(value: boolean | ElementView.InteractivityOptions): void;
 
-        getDelegatedView(): ElementView | null;
+        getDelegatedView(): ElementView<E> | null;
```
kumilingus, line 922: The `getDelegatedView()` function can return a different view (any of the ancestor views).

I think it should stay as it was here.
```ts
getDelegatedView(): ElementView | null;
```

kumilingus (changes requested):
Just this one undocumented method needs to be updated.

Good work!

packages/joint-core/types/joint.d.ts
michaelspiss: Makes sense, fixed

kumilingus (approved):

Issue #2465, opened 2024-01-03
Vectorizer: can not convert zero-length arc to curve


kumilingus:
### Discussed in https://github.com/clientIO/joint/discussions/2464

<div type='discussions-op-text'>

<sup>Originally posted by **gordon-brown** January  3, 2024</sup>
### Introduction

I'm having an issue rendering one of the clip art shapes I converted. It seems to me to be related to rendering elliptical curves in the path data. I can render the shape on the https://www.svgviewer.dev/ site. Not sure if I am doing something wrong or if this is an  issue with your software.
[BPMN_StoredData-202401030954.joint.txt](https://github.com/clientIO/joint/files/13820532/BPMN_StoredData-202401030954.joint.txt)
![Screenshot 2024-01-03 100103](https://github.com/clientIO/joint/assets/32515524/088c92a0-aefe-415b-9b5b-bf24454ee7de)
![Screenshot 2024-01-03 100348](https://github.com/clientIO/joint/assets/32515524/b90d978b-b8c9-48e6-acbc-8ec4f9060e5a)

### Does your question relate to JointJS or JointJS+. Select both if applicable.

JointJS</div>

----

There seems to be an issue with an arc command that has zero length and our code is not able to convert it to the curve command. I am changing the question to a bug.

It works if you define your path like this: 'M 5.555555555555555 0L 50 0A 5.555555555555555,25 0 0,0 50,50 L 50,50L 5.555555555555555 50A 5.555555555555555,25 0 0,1 5.555555555555555,0Z'

#### Visio implications

Also, this is happening when importing joint-format-visio\test\visual\fixtures\Chapter05\Network diagram_start.vsdx.
Path 
``` 
M 0 0 M 32 4 A 3.95 3.95 0 0 1 24.49 3.95 A 3.95 3.95 0 0 1 32.39 3.95 A 3.95 3.95 0 0 1 32.39 3.95
``` 
normalized using `V.normalizePathData` with NaNs as 

```
[...]

PR #2598, opened 2024-03-27
fix(Vectorizer): fix normalizePathData() to support zero-length arc to curve

kumilingus:
## Description

A zero-length arc `A` was converted to an invalid curve command `C` (containing `NaN` values).

After this change, such arcs are converted to a point-like curve.

Fixes https://github.com/clientIO/joint/issues/2465.

### DIFF
--- a/packages/joint-core/src/V/index.mjs
+++ b/packages/joint-core/src/V/index.mjs
@@ -2293,7 +2293,7 @@ const V = (function() {
 
                 var k = ((large_arc_flag == sweep_flag) ? -1 : 1) * sqrt(abs(((rx2 * ry2) - (rx2 * y * y) - (ry2 * x * x)) / ((rx2 * y * y) + (ry2 * x * x))));
                 if (!Number.isFinite(k)) {
-                    // Straight line
+                    // Arc is a single point
                     return [x1, y1, x2, y2, x2, y2];
                 }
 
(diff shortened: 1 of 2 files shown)


packages/joint-core/src/V/index.mjs
```diff
@@ -2292,6 +2292,10 @@ const V = (function() {
                 var ry2 = ry * ry;
 
                 var k = ((large_arc_flag == sweep_flag) ? -1 : 1) * sqrt(abs(((rx2 * ry2) - (rx2 * y * y) - (ry2 * x * x)) / ((rx2 * y * y) + (ry2 * x * x))));
+                if (!Number.isFinite(k)) {
+                    // Straight line
```
zbynekstara, line 2296:
```suggestion
                    // Arc is a single point
```

packages/joint-core/test/vectorizer/vectorizer.js
```diff
@@ -1486,6 +1486,8 @@ QUnit.module('vectorizer', function(hooks) {
             assert.equal(V.normalizePathData('A 3 0 0 0 1 10 15'), 'M 0 0 L 10 15'); // 0 y radius
             assert.equal(V.normalizePathData('A 0 0 0 0 1 10 15'), 'M 0 0 L 10 15'); // 0 x and y radii
 
+            assert.equal(V.normalizePathData('M 3 7 A 7 7 0 0 1 3 7'), 'M 3 7 C 3 7 3 7 3 7'); // straight line
```
zbynekstara, line 1489:
```suggestion
            assert.equal(V.normalizePathData('M 3 7 A 7 7 0 0 1 3 7'), 'M 3 7 C 3 7 3 7 3 7'); // arc corresponding to a single point
```

zbynekstara (approved):

Issue #2648, opened 2024-05-07
[Bug]: Grid-layer display only one when two or more papers exist at the same time


tkaabbc:
### Current versus expected behaviour

Hi team

When I create multiple papers at the same time, I found only one grid-layer of paper can be displayed.  

After all my troubleshooting, I found that the problem was caused by each grid-layers of paper having the same id `pattern_0`.
```html
...
<pattern id="pattern_0"
...
```
```html
<g class="joint-grid-layer" style="pointer-events: none;"><svg xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1" id="v-5" width="0" height="0" x="0" y="0"><defs id="v-4"><pattern id="pattern_0" patternUnits="userSpaceOnUse" width="20" height="20" x="0" y="0"><path id="v-6" d="M 20 0 H0 M0 0 V0 20" stroke="#AAAAAA" stroke-width="1"></path></pattern></defs><rect id="v-8" width="100%" height="100%" fill="url(#pattern_0)"></rect></svg></g>
```
<img width="1668" alt="image" src="https://github.com/clientIO/joint/assets/35325538/bb2852f2-5c09-4603-b634-ca3488bfcf75">

I created a demo:
https://codepen.io/Tingkeng-Wang/pen/VwOZYWy

(And thank you for such a great lib!)

### Steps to reproduce

1. New multiple paper instances at the same time.
2. You will find only one paper display grid layer as expected.

### Version

4.0.1

PR #2654, opened 2024-05-08
fix(dia.Paper): ensure grid pattern ids are unique

kumilingus:
Fixes https://github.com/clientIO/joint/issues/2648

## Description

Fix to generate unique ids for paper `<pattern/>` elements to allow multiple paper instances draw a grid.

By definition, element `id` [must be unique within a single document](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id).

Because the pattern identifiers were unique within the paper instance (and not globally), only one of the grids was displayed.


tkaabbc:
Hi team @kumilingus, may I ask when a new version will be released for this change, I'm currently using JointJS+ and I'm experiencing this problem.

kumilingus:
@tkaabbc we have just released `v4.0.3` patch.

Issue #2328, opened 2023-09-07
[Bug]: Safari doesn't render link arrow markers after loading data


frnora:
### What happened?

Hi,

There is some issue with the markers (not) rendering on Safari, take a look at this image directly from the demo page:
<img width="586" alt="Screen Shot 2023-09-07 at 19 05 01" src="https://github.com/clientIO/joint/assets/129097753/c9e81c9a-aa72-40f9-86ba-43f2f874980b">

The markers appear again when you reconnect the link (or create a new one) but it doesn't work after loading the data. Any ideas?
(Safari Version 16.6)

### Version

3.7

### What browsers are you seeing the problem on?

Safari

### What operating system are you seeing the problem on?

Mac

kumilingus:
When you are using Safari, there is a bug where SVG Markers are not rendered if they are added asynchronously in the next animation frame.

Here is the minimal reproducible case: https://jsfiddle.net/kumilingus/k90h2Lt1/

The issue has been reported. We are currently waiting for the Safari team to address it.

In the meantime, you need to apply a workaround - there are a lot of ways, but it's important to somehow trigger the browser update of the SVG again.
- The issue is not happening when the `async` paper option is set to `false`
- The issue can be resolved for instance by hiding the paper and showing it again (see below).

```js
paper.freeze();
graph.fromJSON(graphJSON);
paper.unfreeze({
    afterRender: function() {
        paper.el.style.display = 'none';
        requestAnimationFrame(() => paper.el.style.display = '');
        paper.unfreeze();
    }
});
```

frnora:
The above code doesn't fire `afterRender` for me, tho. In update `updateViewsAsync` the condition checking for the id of the update evaluates to false (update.id is null). The subsequent calls don't get the `afterRender` fn anymore in the `opt` object.

kumilingus:
If the graph is empty, the `afterRender` callback will not trigger.

You need to check if the graph is empty or check if there are [any updates scheduled](https://github.com/clientIO/joint/issues/1219).

frnora:
Am I misunderstanding this?
Can the graph be empty after `fromJSON` if the json contains cells? I.e.:

```js
paper.freeze();
graph.fromJSON(graphJSON);
console.log(graph.getCells()); // <- correctly logs the cells
paper.unfreeze({
    afterRender: function() {
        paper.el.style.display = 'none';
        requestAnimationFrame(() => paper.el.style.display = '');
        paper.unfreeze();
    }
});
```

[...]

PR #2677, opened 2024-05-30
fix(dia.LinkView): fix missing arrowheads in Safari

kumilingus:
## Description

Safari has a bug where any change after the first render is not reflected in the DOM.
https://bugs.webkit.org/show_bug.cgi?id=268376

This a fix to workaround the issue.

Fixing https://github.com/clientIO/joint/issues/2328.

### DIFF
--- a/packages/joint-core/src/env/index.mjs
+++ b/packages/joint-core/src/env/index.mjs
@@ -9,6 +9,7 @@ export const env = {
                 /SVGForeignObject/.test(({}).toString.call(document.createElementNS('http://www.w3.org/2000/svg', 'foreignObject')));
         },
 
+        // works for iOS browsers too
         isSafari: function() {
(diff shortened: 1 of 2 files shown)


packages/joint-core/src/env/index.mjs
```diff
@@ -7,6 +7,10 @@ export const env = {
         svgforeignobject: function() {
             return !!document.createElementNS &&
                 /SVGForeignObject/.test(({}).toString.call(document.createElementNS('http://www.w3.org/2000/svg', 'foreignObject')));
+        },
+
+        isSafari: function() {
```
zbynekstara, line 12:
```suggestion
        // works for iOS browsers too
        isSafari: function() {
```

zbynekstara (approved):

PR #2723, opened 2024-07-27
feat(linkTools): add Control and RotateLabel link tool

kumilingus:
## Description

TBD - a link tool to rotate labels

image


packages/joint-core/src/elementTools/index.mjs
Geliogabalus, line 1: Is it ok to have export duplicates like this? For compatibility we should have them as it is now, but in the world without namespaces, we should have cellTools package I think.

kumilingus: Currently, it is only an implementation detail that `linkTools.Button` and `elementTools.Button` are the same class (we kind of pretend they are not).
I don't think it breaks any rules.

Geliogabalus (approved):

PR #2728, opened 2024-07-31
feat(dia.Cell): new toJSON options

MartinKanera:
## Description

This PR includes changes to the `dia.Cell.toJSON()`, now accepting two new options.
- `opt.ignoreDefaults` (_default_ `['attrs']`) - accepts an array of attribute names to be compared to the defaults, additionally setting this option to `false` will result in returning all of the attributes without comparing them, while setting it to `true` results in a comparison of all the attributes with defaults
- `opt.ignoreEmptyAttributes` (_default_ `false`) - determines if the result should include empty attributes

These two options are also applicable to `dia.Graph.toJSON()` method.

This PR also includes a new `util.objectDifference` function used to determine the difference between two provided objects.

## Documentation
https://docs.jointjs.com/api/dia/Cell/#tojson

### `toJSON()`

```ts
cell.toJSON([opt])
```

Return a copy of the cell's attributes for JSON serialization. This can be used for persistance or serialization. Note that this method doesn't return a JSON string but rather an object that can be then serialized to JSON with `JSON.stringify()`.

By default, `opt.ignoreDefaults` is set to `['attrs']`. This setting ensures that only non-default attributes in the attrs object are returned, as it compares the attrs object to its defaults. To include all attributes of `dia.Cell` without filtering out defaults, set `opt.ignoreDefaults` to `false`. Alternatively, setting `opt.ignoreDefaults` to `true` will compare all attributes to their defaults, returning only those that differ. You can also provide an array of strings for `opt.ignoreDefaults`, specifying a list of attributes to compare against their defaults.

Additionally, `opt.ignoreEmptyAttributes` defaults to `false`, which may include empty objects in the JSON output. To exclude empty objects, set `opt.ignoreEmptyAttributes` to `true`.

```ts
const rectangle = new shapes.standard.Rectangle();

rectangle.toJSON();
// {
//     "type": "standard.Rectangle",
//     "attrs": {},
//     "position": {
//         "x": 0,
//         "y": 0
//     },
//     "size": {
//         "width": 1,
//         "height": 1
//     },
//     "angle": 0,
//     "id": "..."
// }

rectangle.toJSON({ ignoreDefaults: true });
// {
//     "type": "standard.Rectangle",
//     "attrs": {},
//     "position": {},
//     "size": {},
//     "id": "..."
// }

rectangle.toJSON({ ignoreDefaults: true, ignoreEmptyAttributes: true });
// {
//     "type": "standard.Rectangle",
//     "id": "..."
// }

rectangle.toJSON({ ignoreDefaults: ['attrs', 'size'] });
// {
//     "type": "standard.Rectangle",
//     "attrs": {},
//     "position": {
//         "x": 0,
//         "y": 0
//     },
//     "size": {},
//     "angle": 0,
//     "id": "..."
// }
```
---
https://docs.jointjs.com/api/dia/Graph/#tojson

### toJSON()

```ts
graph.toJSON([opt])
```

This method accepts an optional `opt` object, the `opt.cellAttributes` accepts same options as the [dia.Cell.toJSON()](../Cell/Cell.mdx#tojson) method, which will impact the result of the `toJSON()` method.

---
https://docs.jointjs.com/api/util/

### objectDifference()

```ts
util.objectDifference(object, base, [opt])
```

Return an object that represents the difference between `object` and `base`.

<table>
    <tbody>
        <tr>
            <th>object</th>
            <td>Object</td>
            <td>The object to compare</td>
        </tr>
        <tr>
            <th>base</th>
            <td>Object</td>
            <td>The object to compare with</td>
        </tr>
        <tr>
            <th>[opt.maxDepth]</th>
            <td>number</td>
            <td>The maximum depth to compare</td>
        </tr>
    </tbody>
</table>

### DIFF
--- a/packages/joint-core/src/dia/Cell.mjs
+++ b/packages/joint-core/src/dia/Cell.mjs
@@ -43,6 +43,23 @@ const attributesMerger = function(a, b) {
     }
 };
 
+function removeEmptyAttributes(obj) {
+
+    for (const key in obj) {
+
+        const objValue = obj[key];
+        const isRealObject = isObject(objValue) && !Array.isArray(objValue);
+
+        if (isRealObject) {
+            removeEmptyAttributes(objValue);
+        }
+
+        if (isEmpty(objValue) && isRealObject) {
+            delete obj[key];
+        }
+    }
+}
+
 export const Cell = Model.extend({
 
     // This is the same as mvc.Model with the only difference that is uses util.merge
@@ -81,23 +98,6 @@ export const Cell = Model.extend({
         const { ignoreDefaults, ignoreEmptyAttributes = false } = opt || {};
         const defaults = result(this.constructor.prototype, 'defaults');
 
-        function removeEmptyAttributes(obj) {
-
-            for (const key in obj) {
-
-                const objValue = obj[key];
-                const isRealObject = isObject(objValue) && !Array.isArray(objValue);
-
-                if (isRealObject) {
-                    removeEmptyAttributes(objValue);
-                }
-
-                if (isEmpty(objValue) && isRealObject) {
-                    delete obj[key];
-                }
-            }
-        }
-
         if (ignoreDefaults === false) {
             // Return all attributes without omitting the defaults
             const finalAttributes = cloneDeep(this.attributes);
@@ -125,7 +125,7 @@ export const Cell = Model.extend({
         }
 
         // Omit `type` attribute from the defaults since it should be always present
-        const finalAttributes = objectDifference(attributes, omit(defaultAttributes, 'type'), ignoreEmptyAttributes, 4);
+        const finalAttributes = objectDifference(attributes, omit(defaultAttributes, 'id', 'type'), ignoreEmptyAttributes, 4);
 
         return finalAttributes;
     },
--- a/packages/joint-core/src/dia/Cell.mjs
+++ b/packages/joint-core/src/dia/Cell.mjs
@@ -25,7 +25,8 @@ import {
     defaultsDeep,
     has,
     sortBy,
-    defaults
+    defaults,
+    objectDifference
 } from '../util/util.mjs';
 import { Model } from '../mvc/Model.mjs';
 import { cloneCells } from '../util/cloneCells.mjs';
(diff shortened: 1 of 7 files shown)


packages/joint-core/types/joint.d.ts
```diff
@@ -239,7 +239,7 @@ export namespace dia {
 
         getCommonAncestor(...cells: Cell[]): Element | undefined;
 
-        toJSON(): any;
+        toJSON(opt?: { ignoreDefaults: boolean | string[], ignoreEmptyAttributes: boolean }): any;
```
kumilingus, line 242: Can we create `dia.Cell.ExportOptions` type and use it for the graph method too?

packages/joint-core/src/util/util.mjs
```diff
@@ -1708,6 +1708,47 @@ export const toggleFullScreen = function(el) {
     }
 };
 
+
+// Util function used for purposes of `dia.Cell.toJSON()` method
+export function objectDifference(object, base, ignoreEmptyObjects = false, depth = Number.POSITIVE_INFINITY) {
+
+    function findDifference(obj, baseObj, currentDepth) {
```
kumilingus, line 1715: Can we define it outside the exported function (+ add `ignoreEmptyObjects` object)?

packages/joint-core/src/dia/Cell.mjs
```diff
@@ -75,48 +76,58 @@ export const Cell = Model.extend({
...
+        const { ignoreDefaults, ignoreEmptyAttributes = false } = opt || {};
         const defaults = result(this.constructor.prototype, 'defaults');
-        const defaultAttrs = defaults.attrs || {};
-        const attrs = this.attributes.attrs;
-        const finalAttrs = {};
 
-        // Loop through all the attributes and
-        // omit the default attributes as they are implicitly reconstructible by the cell 'type'.
-        forIn(attrs, function(attr, selector) {
+        function removeEmptyAttributes(obj) {
```
kumilingus, line 84: Can we define the function in the outer scope?

packages/joint-core/src/dia/Cell.mjs
```diff
@@ -75,48 +76,58 @@ export const Cell = Model.extend({
...
+            differentiateKeys.forEach((key) => {
+                defaultAttributes[key] = defaults[key] || {};
             });
-        });
+        }
 
-        const attributes = cloneDeep(omit(this.attributes, 'attrs'));
-        attributes.attrs = finalAttrs;
+        // Omit `type` attribute from the defaults since it should be always present
+        const finalAttributes = objectDifference(attributes, omit(defaultAttributes, 'type'), ignoreEmptyAttributes, 4);
```
kumilingus, line 128: I would change the signature to `objectDifference(obj, base, { ignoreEmptyAttributes, maxDepth });` and make the function public.

Also does `ignoreEmptyAttributes` ignore the nested attributes too?

kumilingus: We should omit `id` attribute too, to be safe.

kumilingus (changes requested):

kumilingus (approved):

PR #2735, opened 2024-08-10
feat(dia.CellView): add unset() attribute callback

kumilingus:
## Description

If you currently set `sourceMarker` and later try to remove it by setting it to `null`, `marker-start` SVG attribute will not be removed from the DOM. This is a mechanism to ensure that the actual presentation attributes can be removed if necessary.

This PR also defines TypeScript for defining custom special attributes and fixes various examples where special attributes were used.

### Documentation

The `unset` callback for presentation attributes is called when an attribute is set to `null`.
- If there is no `unset` callback defined, an attribute with the same name as attribute definition is removed.
- Otherwise, it can be a **string** (an attribute to be removed from the DOM element), an **array of strings**, or a function that returns the aforementioned string or array.
- The callback accepts `(node, attrs, cellView)`.

```ts
attributes: {
  'line-style': {
      set: function(lineStyle, refBBox, node, attrs) {
          const n = parseFloat(attrs['stroke-width']) || 1;
          const dasharray = {
              'dashed': `${4*n},${2*n}`,
              'dotted': `${n},${n}`,
          }[lineStyle] || 'none';
          return { 'stroke-dasharray': dasharray };
      },
      unset: 'stroke-dasharray'
  }
}
```

### DIFF
--- a/packages/joint-core/demo/custom-shapes/src/custom-shapes.mjs
+++ b/packages/joint-core/demo/custom-shapes/src/custom-shapes.mjs
@@ -24,7 +24,8 @@ joint.dia.attributes['line-style'] = {
         }[lineStyle] || 'none';
 
         return { 'stroke-dasharray': dasharray };
-    }
+    },
+    unset: 'stroke-dasharray'
 };
 
 joint.dia.attributes['fit-ref'] = {
@@ -49,7 +50,8 @@ joint.dia.attributes['fit-ref'] = {
                 };
         }
         return {};
-    }
+    },
+    unset: ['rx', 'ry', 'cx', 'cy', 'width', 'height', 'd']
 };
 
 var Circle = joint.dia.Element.define('custom.Circle', {
--- a/packages/joint-core/demo/custom-shapes/src/custom-shapes.mjs
+++ b/packages/joint-core/demo/custom-shapes/src/custom-shapes.mjs
@@ -2,9 +2,9 @@ import * as joint from '../../../joint.mjs';
 import * as g from '../../../src/g/index.mjs';
 import V from '../../../src/V/index.mjs';
 
-var graph = new joint.dia.Graph({}, { cellNamespace: joint.shapes });
+const graph = new joint.dia.Graph({}, { cellNamespace: joint.shapes });
 
-var paper = new joint.dia.Paper({
+const paper = new joint.dia.Paper({
     el: document.getElementById('paper'),
     cellViewNamespace: joint.shapes,
     width: 650,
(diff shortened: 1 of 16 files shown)


packages/joint-core/demo/custom-shapes/src/custom-shapes.mjs
zbynekstara, line 1: What about the other custom attributes set in this file (Shape > shape sets 'd' and Progress > progress-d sets 's') - should `unset` be added for them too?

zbynekstara, line 1: Also, a few other demos define shapes with custom presentation attributes (puzzle, rough, standard) - what about them?

kumilingus: We should update them too. Will do.

packages/joint-core/src/dia/attributes/connection.mjs
zbynekstara, line 1: Should we add `unset` for custom presentation attributes using `atConnectionWrapper`? They modify `transform`

packages/joint-core/src/dia/attributes/index.mjs
zbynekstara, line 1: Should we add `unset` for 'style'? I guess the styles can already be unset by providing an empty object tough

packages/joint-core/src/dia/attributes/shape.mjs
zbynekstara, line 1: Should we do a similar thing for the `ref` attributes inside `src/dia/attributes/legacy.mjs` too? They are setting `x`, `y`, `width`, `height`, `rx`, `ry`, `cx`, `cy`, `r`

kumilingus: no, legacy attributes are deprecated and no longer maintained.

packages/joint-core/src/dia/attributes/index.mjs
kumilingus: The `style` special attribute sets the `style` HTML/SVG attribute to the elements. So there is no need to explicitly state this.

packages/joint-core/src/dia/attributes/connection.mjs
kumilingus: Yes. I've added a test to check whether it still work fine with the `position` attribute callback.

packages/joint-core/types/joint.d.ts
```diff
@@ -339,6 +339,46 @@ export namespace dia {
...
+            node: Element,
+            nodeAttributes: { [name: string]: any },
+            cellView: V
+        ) => string | Array<string> | null | void;
+
+        type SetCallback<V> = (
+            this: V,
+            attributeValue: any,
+            refBBox: g.Rect,
+            nodeAttributes: { [name: string]: any },
```
zbynekstara, line 354:
```suggestion
            node: Element,
            nodeAttributes: { [name: string]: any },
```

packages/joint-core/types/joint.d.ts
```diff
@@ -339,6 +339,46 @@ export namespace dia {
...
+            cellView: V
+        ) => { [key: string]: any } | string | number;
```
zbynekstara, line 356:
```suggestion
        ) => { [key: string]: any } | string | number | void;
```

zbynekstara (approved):

PR #2752, opened 2024-09-06
feat(dia.Graph): add transferCellEmbeds() and transferCellConnectedLinks()

MartinKanera:
## Description

Add `graph.transferCellEmbeds()` and `graph.transferCellConnectedLinks()`.
Change the `dia.Cell.embed` method to accept `opt.reparent`.

### transferCellEmbeds()

```ts
graph.transferCellEmbeds(sourceCell, targetCell [, opt]);
```

Transfer embedded cells of `sourceCell` to `targetCell`.

### transferCellConnectedLinks()

```ts
graph.transferCellConnectedLinks(sourceCell, targetCell [, opt]);
```

Transfer all the links connected to `sourceCell` to `targetCell`.

### embed()

...

The `opt.reparent` option can be set to `true` to transfer the embedded cells from their current parent to the new parent.

### DIFF
--- a/packages/joint-core/src/dia/Graph.mjs
+++ b/packages/joint-core/src/dia/Graph.mjs
@@ -398,6 +398,8 @@ export const Graph = Model.extend({
 
     transferCellHierarchy: function(sourceCell, targetCell) {
 
+        this.startBatch('transfer-hierarchy');
+
         // Embed children of the source cell in the target cell.
         const children = sourceCell.getEmbeddedCells();
         if (children.length > 0) {
@@ -412,10 +414,14 @@ export const Graph = Model.extend({
             parent.embed(targetCell);
         }
 
+        this.stopBatch('transfer-hierarchy');
+
         return this;
     },
 
     transferCellLinks: function(sourceCell, targetCell) {
+
+        this.startBatch('transfer-links');
 
         // Reconnect all the links connected to the old cell to the new cell.
         const connectedLinks = this.getConnectedLinks(sourceCell);
@@ -429,6 +435,8 @@ export const Graph = Model.extend({
                 link.prop(['target', 'id'], targetCell.id);
             }
         });
+
+        this.stopBatch('transfer-links');
 
         return this;
     },
--- a/packages/joint-core/src/dia/Graph.mjs
+++ b/packages/joint-core/src/dia/Graph.mjs
@@ -396,6 +396,41 @@ export const Graph = Model.extend({
         this.get('cells').remove(cell, { silent: true });
     },
 
+    transferCellHierarchy: function(sourceCell, targetCell) {
+
+        // Embed children of the source cell in the target cell.
+        const children = sourceCell.getEmbeddedCells();
+        if (children.length > 0) {
+            sourceCell.unembed(children);
+            targetCell.embed(children);
+        }
+
+        // Embed the target cell in the parent of the source cell, if any.
+        const parent = sourceCell.getParentCell();
+        if (parent) {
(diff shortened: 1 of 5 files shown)


packages/joint-core/src/dia/Graph.mjs
```diff
@@ -396,6 +396,41 @@ export const Graph = Model.extend({
...
+        return this;
+    },
+
+    transferCellLinks: function(sourceCell, targetCell) {
+
+        // Reconnect all the links connected to the old cell to the new cell.
+        const connectedLinks = this.getConnectedLinks(sourceCell);
+        connectedLinks.forEach((link) => {
+
+            if (link.getSourceCell() === sourceCell) {
```
kumilingus, line 424: I don't think logic will work for loop links i.e the same source and target

kumilingus (commented):
Both  methods should run inside a batch:
`transfer-links`
`transfer-hiearchy`

zbynekstara:
Thoughts on the naming of the functions:
- `graph.transferCellHierarchy` is not ideal because it makes it seem that target cell will not be modified (only passively receiving the children of source element), but I have trouble coming up with a better name. I don't think the word `cellHierarchy` makes it obvious that target will also be reparented to source's parent. I don't think these two separate actions can be captured by a single word - perhaps the name should be split to describe the two actions separately and more precisely, like `graph.transferCellEmbedsAndReparent`?
- `graph.transferCellLinks` is okay, but since we already have `cell.getConnectedLinks`, `graph.transferCellConnectedLinks` would be more consistent

packages/joint-core/types/joint.d.ts
```diff
@@ -263,6 +263,10 @@ export namespace dia {
 
         removeCells(cells: Cell[], opt?: Cell.DisconnectableOptions): this;
 
+        transferCellEmbeds(sourceCell: Cell, targetCell: Cell, opt?: { [key: string]: any }): void;
+
+        transferCellConnectedLinks(sourceCell: Cell, targetCell: Cell, opt?: { [key: string]: any }): void;
```
kumilingus, line 268: The `opt` here is a `Graph.ConnectionOptions`. You can for instance, reconnect only the `inbound` links.

packages/joint-core/src/dia/Graph.mjs
```diff
@@ -396,6 +396,41 @@ export const Graph = Model.extend({
...
 
+    transferCellEmbeds: function(sourceCell, targetCell, opt = {}) {
+
+        const batchName = 'transfer-embeds';
+        this.startBatch(batchName);
+
+        opt.reparent = true;
+
+        // Embed children of the source cell in the target cell.
+        const children = sourceCell.getEmbeddedCells(opt);
```
kumilingus, line 407: Why do we pass `opt` here? Do we want `deep: true` to be use here?
That would work with `{ deep: true, reparent: true }` but I am not sure if this is a scenario we want to support.

packages/joint-core/src/dia/Graph.mjs
```diff
@@ -396,6 +396,41 @@ export const Graph = Model.extend({
         this.get('cells').remove(cell, { silent: true });
     },
 
+    transferCellEmbeds: function(sourceCell, targetCell, opt = {}) {
+
+        const batchName = 'transfer-embeds';
+        this.startBatch(batchName);
+
+        opt.reparent = true;
```
kumilingus, line 404: I would not modify the `opt`. It would be better to pass `{ ...opt, reparent: true }` inside the `embed()` function.

kumilingus (approved):

PR #2781, opened 2024-10-27
feat(dia.Paper):  add methods to find cell/element/link views in paper

kumilingus:
- **feat(dia.Paper):**  add methods to find cell/element/link views in paper
  - add `findCellViewsAtPoint(), `findLinkViewsAtPoint()`, and `findElementViewsAtPoint()`
  - add `findCellViewsInArea(), `findLinkViewsInArea()`, and `findElementViewsInArea()`
  - deprecate `findViewsFromPoint()` and `findViewsInArea()
- **feat(dia.Graph):**  add methods to find cells/elements/links in graph
  - add `findCellsAtPoint()`, `findLinksAtPoint()`, and `findElementsAtPoint()`
  - add `findCellsInArea()`, `findLinksInArea()`, and `findElementsInArea()`
  - add `findCellsUnderElement()`, `findLinksUnderElement()`, `findElementsUnderElement()`
  - deprecate `findModelsFromPoint()` and `findModelsInArea()`
- **feat(dia.CellView):** add `isIntersecting()` method
- **feat(g.Rect):** add `strict` option to `containsPoint()`

---

This PR:
- adds convenience methods for searching for links in the paper based on their geometry.
- optimizes the performance of search for elements and links in the paper.
- allows us to further optimize the search by a more efficient data structure in the graph.
- It lays the foundation for the future implementation of `snap-to-link` feature (as opposed to existing `snap-to-elements`).

### DIFF
--- a/packages/joint-core/src/dia/LinkView.mjs
+++ b/packages/joint-core/src/dia/LinkView.mjs
@@ -817,14 +817,14 @@ export const LinkView = CellView.extend({
         return connectionPoint.round(this.decimalsRounding);
     },
 
-    isIntersecting: function(geometryShape, options) {
+    isIntersecting: function(geometryShape, geometryData) {
         const connection = this.getConnection();
         if (!connection) return false;
         return intersection.exists(
-            connection,
             geometryShape,
+            connection,
+            geometryData,
             { segmentSubdivisions: this.getConnectionSubdivisions() },
-            options
         );
     },
 
--- a/packages/joint-core/src/dia/LinkView.mjs
+++ b/packages/joint-core/src/dia/LinkView.mjs
@@ -73,6 +73,7 @@ export const LinkView = CellView.extend({
     initFlag: [Flags.RENDER, Flags.SOURCE, Flags.TARGET, Flags.TOOLS],
 
     UPDATE_PRIORITY: 1,
+    EPSILON: 1e-6,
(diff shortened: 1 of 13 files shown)


packages/joint-core/src/dia/LinkView.mjs
```diff
@@ -816,6 +817,34 @@ export const LinkView = CellView.extend({
         return connectionPoint.round(this.decimalsRounding);
     },
 
+    isIntersecting: function(geometryShape, options) {
```
Geliogabalus, line 820: As I understand, here it is different signature than in CellView (options here goes into 4th parameter of intersection call). I think it should has the same signature as CellView because it is effectively overrides parent method. And it is not reflected in types.

kumilingus: It's actually ok.

```ts
return intersection.exists(
    connection,
    geometryShape,
    { segmentSubdivisions: this.getConnectionSubdivisions() },
    options
);
```

and

```ts
return intersection.exists(
    geometryShape,
    connection,
    options,
    { segmentSubdivisions: this.getConnectionSubdivisions() },
);
```

are equivalent.

Let me at least rename the `options` param to `geometryData` to match the parent class.

Geliogabalus (approved):

PR #2783, opened 2024-10-30
feat(layout.DirectedGraph): add `graph` option to fromGraphLib()

kumilingus:
## Description

The `fromGraphLib()` utility can now be used without providing the graph reference through context (this was an oversight during the conversion of `DirectedGraph` into a standalone package).

For backward compatibility, passing graph through context is still supported but is now deprecated. Instead, the graph should be provided via the `graph` option.

### DIFF
--- a/packages/joint-layout-directed-graph/DirectedGraph.mjs
+++ b/packages/joint-layout-directed-graph/DirectedGraph.mjs
@@ -34,10 +34,10 @@ export const DirectedGraph = {
         return edge;
     },
 
-    importElement: function(opt, v, gl) {
+    importElement: function(nodeId, glGraph, graph, opt) {
 
-        var element = this.getCell(v);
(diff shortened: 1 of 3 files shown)


packages/joint-layout-directed-graph/DirectedGraph.mjs
```diff
@@ -49,12 +49,12 @@ export const DirectedGraph = {
         }
     },
 
-    importLink: function(opt, edgeObj, gl) {
+    importLink: function(edgeObj, glGraph, graph, opt) {
 
         const SIMPLIFY_THRESHOLD = 0.001;
 
         const link = this.getCell(edgeObj.name);
```
zbynekstara, line 56:
```suggestion
        const link = graph.getCell(edgeObj.name);
```

zbynekstara (approved):

PR #2839, opened 2024-12-19
feat(dia.LinkView): allow link to link snap

MartinKanera:
## Description

Enhance link snapping functionality: When `paper.options.snapLinks = true`, update the snapping logic to consider all cells, not just Elements. This ensures that links can snap to any cell within the paper.

PR also fixes cases where the link snapped to a magnet that was not within the radius (only the entire element was within the radius).

### DIFF
--- a/packages/joint-core/src/dia/ElementView.mjs
+++ b/packages/joint-core/src/dia/ElementView.mjs
@@ -504,6 +504,10 @@ export const ElementView = CellView.extend({
         });
     },
 
+    getSnapPoint: function() {
+        return this.model.getBBox().center();
+    },
+
     _disallowUnembed: function(data) {
         const { model, whenNotAllowed = 'revert' } = data;
         const element = model || this.model;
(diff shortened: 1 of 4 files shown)


packages/joint-core/src/dia/LinkView.mjs
```diff
@@ -1873,27 +1873,46 @@ export const LinkView = CellView.extend({
...
+                        distance: model.getBBox().center().squaredDistance(pointer),
+                        magnet: view.el
+                    });
+                }
             }
 
             view.$('[magnet]').toArray().forEach(magnet => {
+
+                const magnetBBox = view.getNodeBBox(magnet);
+                const rootCenter = model.isElement() ? model.getBBox().center() : view.getConnection().pointAt(0.5);
```
kumilingus, line 1902: This should be moved to `elementView` and `linkView`.  
For example, you could add a `getSnapPoint()` method to allow overriding the logic.

packages/joint-core/src/dia/ElementView.mjs
```diff
@@ -504,6 +504,10 @@ export const ElementView = CellView.extend({
         });
     },
 
+    getSnapPoint: function() {
```
kumilingus: I don't think this is being used.

kumilingus (approved):

PR #2844, opened 2025-01-02
feat(dia.Paper): layers API

kumilingus:
## Description

An API to add / remove / change stacking order of paper layers.


packages/joint-core/src/dia/Paper.mjs
```diff
@@ -1796,8 +1905,11 @@ export const Paper = View.extend({
     },
 
     insertView: function(view, isInitialInsert) {
-        const layerView = this.getLayerView(LayersNames.CELLS);
         const { el, model } = view;
+
+        const layerName = model.get('layer') || this.DEFAULT_CELL_LAYER;
```
Geliogabalus, line 1910: This attribute should be reflected in types alongside `z`.

Geliogabalus (approved):

PR #2879, opened 2025-02-17
feat(dia.Element): add filter and minRect options to fitToChildren() and fitParent()

zbynekstara:
## Description

Provides two new options for `element.fitToChildren()` and `element.fitParent()` methods:
- `opt.filter` = provide a callback function that returns `true` when an element-type child should be included
- `opt.minRect` = provide an object with `x`, `y`, `width` and `height` which specifies a minimum rectangle that must be part of the bounding box after fitting (added via union with the fitting result)

Edge cases:
- `opt.padding` is not applied on top of `opt.minRect` because `opt.padding` only specifies padding that embedded elements must have from the bounding box after fitting and thus doesn't apply to empty space inside the minimal rectangle
- When used together with `opt.deep`, `opt.minRect` is only applied for the first iteration, but not for any other
  - If the element has no filtered children, then `fitToChildren()` applies `opt.minRect` to the element; this is not the case for further descendants.
  - If the element has no filtered siblings, then `fitParent()` applies `opt.minRect` to the parent; this is not the case for further ancestors.
- If there is no parent, then `fitParent()` exits early without considering `opt.minRect` (since there is nothing to apply it to).
- In all other cases where we previously exited early, `opt.minRect` is now applied instead as a last resort (if it is provided):
  - if there are no filtered elements
  - if `opt.expandOnly` and `opt.shrinkOnly` are used together
  - if `opt.shrinkOnly` is in use and all children are outside the element's bounding box

## Motivation and Context

The `filter` option gives you full control over what kinds of embedded elements should be included in fitting. This is useful when you implement additional control logic that affects the contents of parent elements (e.g. when some child elements become invisible).

The `minRect` option gives you an opportunity to specify a fallback minimal bounding box for fitting. This may be necessary when your container needs to show placeholder text, or in general, when your container resizes according to contents but must stay prominently visible when there is not enough content for it.

The following example combines the `filter` option with the `padding` and `minRect` options. The `minRect` option makes sure that the element in question (the container) has a minimal width and height so that it is always visible to users (even if it only contains one small element, or if all contained elements are filtered out), while the `padding` option ensures that the container provides enough visual space around elements when some are present:

```ts
fitToChildElements: function (flags) {
    const HEADER_HEIGHT = 30;
    const PADDING = 10;
    const MIN_WIDTH = 140;
    const MIN_HEIGHT = 100;
    const origin = this.getBBox().origin();
    this.fitToChildren(
        {
            filter: (child) => (!this.isChildFiltered(child)),
            padding: {
                top: HEADER_HEIGHT + PADDING,
                left: PADDING,
                right: PADDING,
                bottom: PADDING
            },
            minRect: { x: origin.x, y: origin.y, width: MIN_WIDTH, height: MIN_HEIGHT },
        },
        flags
      );
    }
```

### DIFF
--- a/packages/joint-core/src/dia/Element.mjs
+++ b/packages/joint-core/src/dia/Element.mjs
@@ -345,8 +345,12 @@ export const Element = Cell.extend({
         if (!graph) throw new Error('Element must be part of a graph.');
 
         // Get filtered element children.
-        let filterFn = (cell) => (cell.isElement());
-        if (typeof opt.filter === 'function') filterFn = (cell) => (cell.isElement() && opt.filter(cell));
+        let filterFn;
+        if (typeof opt.filter === 'function') {
+            filterFn = (cell) => (cell.isElement() && opt.filter(cell));
+        } else {
+            filterFn = (cell) => (cell.isElement());
+        }
         const childElements = this.getEmbeddedCells().filter(filterFn);
         if (childElements.length === 0) return this;
 
@@ -383,8 +387,12 @@ export const Element = Cell.extend({
         if (!parentElement || !parentElement.isElement()) return this;
 
         // Get filtered element children of parent element (i.e. this element + any sibling elements).
-        let filterFn = (cell) => (cell.isElement());
-        if (typeof opt.filter === 'function') filterFn = (cell) => (cell.isElement() && opt.filter(cell));
+        let filterFn;
+        if (typeof opt.filter === 'function') {
+            filterFn = (cell) => (cell.isElement() && opt.filter(cell));
+        } else {
+            filterFn = (cell) => (cell.isElement());
+        }
         const siblingElements = parentElement.getEmbeddedCells().filter(filterFn);
         if (siblingElements.length === 0) return this;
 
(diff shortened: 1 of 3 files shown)


packages/joint-core/src/dia/Element.mjs
```diff
@@ -357,9 +360,18 @@ export const Element = Cell.extend({
         }
 
         // Set new size and position of this element, based on:
-        // - union of bboxes of all children
+        // - union of bboxes of filtered element children
         // - inflated by given `opt.padding`
+        // - containing at least `opt.minRect`
         this._fitToElements(Object.assign({ elements: childElements }, opt));
+        if (opt.minRect) {
```
kumilingus, line 367: why is this logic here (and duplicated) and no in the `_fitToElements()` function?

The function also currently sets `position` and `size` twice.

kumilingus:
is `padding` added to the `minRect` rectangle?
```ts
// assert: el.getEmbeddedChildren().length === 0
el.fitToChildren({ minRect: { x: 10, y: 10, width: 80, height: 80 }, padding: 10 });
// el.getBBox() ??
```

kumilingus (approved):

PR #2908, opened 2025-03-18
feat(layers): Graph and Paper embedding layers implementation

Geliogabalus:
## Description

Adds new functionality to the Graph and Paper which allows to use layers for the embedding. The behaviour controlled by `useLayersForEmbedding` graph option. Work in progress.

## Motivation and Context

Make possible to use layers for embedding.


packages/joint-core/src/dia/Graph.mjs
```diff
@@ -70,6 +71,17 @@ export const Graph = Model.extend({
 
         opt = opt || {};
 
+        this.enableCellLayers = opt.enableCellLayers || false;
+
+        this.defaultLayerName = 'cells';
```
kumilingus: we should move this outside as class property

packages/joint-core/src/dia/Graph.mjs
```diff
@@ -70,6 +71,17 @@ export const Graph = Model.extend({
 
         opt = opt || {};
 
+        this.enableCellLayers = opt.enableCellLayers || false;
```
kumilingus: does not seem to be used anywhere

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -617,91 +628,109 @@ export const Paper = View.extend({
...
     },
 
     moveLayer(layer, insertBefore) {
-        const layerView = this._requireLayerView(layer);
-        if (layerView === this._getLayerView(insertBefore)) return;
-        const layerName = this._getLayerName(layerView);
-        this._unregisterLayer(layerView);
-        this.addLayer(layerName, layerView, { insertBefore });
+        if (!layer) {
+            throw new Error('dia.Paper: The layer view is not an instance of dia.LayerView.');
```
kumilingus: the error is not exactly what the check does

packages/joint-core/src/dia/Graph.mjs
```diff
@@ -429,35 +414,69 @@ export const Graph = Model.extend({
...
     getFirstCell: function() {
 
-        return this.get('cells').first();
+        const cells = this.getCells();
+
+        return cells[0];
     },
 
+    // @deprecated
     getLastCell: function() {
```
kumilingus: should accept `layerName` now and not be deprecated

packages/joint-core/src/dia/HighlighterView.mjs
```diff
@@ -117,7 +117,7 @@ export const HighlighterView = mvc.View.extend({
                 vGroup = V('g').addClass('highlight-transform').append(el);
             }
             this.transformGroup = vGroup;
-            paper.getLayerView(layerName).insertSortedNode(vGroup.node, options.z);
+            paper.getLayer(layerName).insertSortedNode(vGroup.node, options.z);
```
kumilingus: `paper.geLayer()` should stay `paper.getLayerView()` (same analogy to `graph.findCellsInArea()` and `paper.findCellViewsInArea()`

PR #2932, opened 2025-04-14
feat(layout-directed-graph): support clusterPadding: 'default'

zbynekstara:
## Description

Supports `opt.clusterPadding: 'default'`, which applies default Dagre padding logic for clustered elements.

The option is only relevant when `opt.resizeClusters: true` (this was already the case when `opt.clusterPadding` is a number).

429782922-8a0d5990-2666-4e0d-a004-3eb5594741f8


kumilingus (approved):

kumilingus (changes requested):
The TypeScript definition needs to be updated.

kumilingus (approved):

PR #2936, opened 2025-04-17
feat(joint-react): text-node + add some missing stories tests + jest …

samuelgja:
https://github.com/orgs/clientIO/projects/6/views/13?pane=issue&itemId=105648140
## Description
- Update jest.setup to match latest config - it fix some issues with snapshots
- Adding new joint-react `TextNode` component
- Change alias for root storybook folder - so we can have storybook snapshot tests as well
- Writing some another storybook tests

## Motivation and Context
Jointjs has textNode utilities, so we use them for svg Text - now in JointReact there is new `TextNode` component. 

Plus adding snapshots and fix typescript alias for storybook.

### Screenshots:

https://github.com/user-attachments/assets/e0f6da9e-46fb-4f9d-babd-71355476c932


packages/joint-react/src/components/text-node/text-node.tsx
```diff
@@ -0,0 +1,120 @@
...
+    lineHeight,
+    maxLineCount,
+    preserveSpaces,
+    separator,
+    textPath,
+    textVerticalAnchor,
+    width,
+    x,
+  ]);
+  return <text ref={textRef} {...rest} width={width} height={height} x={x} />;
```
kumilingus, line 119: there is no `width` and `height` attribute for `<text/>`, right?

samuelgja: there is intelisense for it, so I thought it was there..I removed as from specs it do nothing.
Screenshot 2025-04-18 at 11 33 09

PR #2972, opened 2025-05-23
feat(examples/container): move to examples, convert to TS, render collapse button via highlighter

zbynekstara:
## Description

Converts `joint/demo/container` into TS and moves to `examples/container`.

Renders containers' collapse/expand buttons with a highlighter, instead of hard-coding them in their markup.

## Motivation and Context

Using highlighters for this purpose is a better practice, since they can be mixed-and-matched with any shape.


packages/joint-core/types/joint.d.ts
```diff
@@ -3486,7 +3486,7 @@ export namespace mvc {
 
         /**
          * Events hash or a method returning the events hash that maps events/selectors to methods on your View.
-         * For assigning events as object hash, do it like this: this.events = <any>{ "event:selector": callback, ... };
+         * For assigning events as object hash, do it like this: `this.events = <any>{ "event:selector": callback, ... };`
```
kumilingus, line 3489: why is this here?

examples/container/src/shapes.ts
```diff
@@ -0,0 +1,195 @@
...
+        return {
+            ...super.defaults,
+            type: 'container.Child',
+            size: { width: 50, height: 50 },
+            attrs: {
+                root: {
+                    magnetSelector: 'body'
+                },
+                shadow: {
+                    refWidth: '100%',
```
kumilingus, line 48: change it to `calc()` expression please

packages/joint-core/types/joint.d.ts
zbynekstara: This JSDoc was relevant for me when doing the conversion to TS, and the missing `` broke the formatting of the code example so that it did not make sense, so I fixed it.

I can remove it from this PR if you want

PR #2974, opened 2025-05-26
dia.Paper: add `viewManagement` option

kumilingus:
- **feat(dia.Paper)** - add a new `viewManagement` option to `dia.Paper` that enables advanced view lifecycle management with lazy initialization and optional disposal of hidden views. The primary purpose is to improve performance in scenarios with large numbers of cells by deferring view creation and potentially disposing views that are not visible.
  - Implementation of a placeholder system for lazy view initialization
  - Introduction of the `viewManagement` option with `lazyInitialize` and `disposeHidden` sub-options
  - Replacement of the `viewport` callback with a new `cellVisibility` callback to support lazy view initialization

- **feat(dia.HighlighterView)** - add static `has()` method to check if a cell view has a highlighter attached

- **feat(mvc.View)** - custom `cid` can be provided via constructor

- **demo** - update `async` demo with the `viewManagement` option


packages/joint-core/src/dia/Paper.mjs
```diff
@@ -1759,59 +1830,115 @@ export const Paper = View.extend({
         return restrictedArea;
     },
 
-    createViewForModel: function(cell) {
+    _resolveCellViewPlaceholder: function(placeholder) {
+        const { model, viewClass, cid } = placeholder;
+        const view = this._initializeCellView(viewClass, model, cid);
+        view.paper = this;
+        this._views[model.id] = view;
```
Geliogabalus: This part can be converted into the separate function or just put into the `createViewForModel` method. There is very similar part in the `renderView` method (line 1972)

packages/joint-core/types/joint.d.ts
```diff
@@ -1992,7 +1992,9 @@ export namespace dia {
 
         protected insertView(cellView: CellView, isInitialInsert: boolean): void;
 
-        protected detachView(cellView: CellView): void;
+        protected _hideCellView(cellView: CellView): void;
```
Geliogabalus, line 2010: The naming is different from all other methods we have currently

packages/joint-core/src/dia/Paper.mjs
kumilingus: Great point!

packages/joint-core/types/joint.d.ts
kumilingus: There are two reasons I chose this name:
- Some methods are designed to work with any `mvc.View` (such as requestViewUpdate()), while others apply specifically to `CellView`. It’s important to distinguish between these cases clearly.
- I use the underscore (`_`) to indicate that this method is not public, and also that it should never be called independently, it only hides the view without updating internal structures.

packages/joint-core/src/dia/Paper.mjs
kumilingus: I've decided to move the logic into a separate `_registerCellView()` method.
The purpose of `createViewForModel` should remain unchanged i.e. it should create a view and allow the user to decide how to use it. The view doesn’t necessarily have to be used with this paper.

packages/joint-core/types/joint.d.ts
Geliogabalus: Ok, I think if it is not supposed to be called, then underscore is fine in this case. We have more untyped methods with underscore currently, we can expose them later in the same vein.

packages/joint-core/src/dia/Paper.mjs
Geliogabalus: Nice, it is more clear what is going on now.

Geliogabalus (approved):

PR #2979, opened 2025-05-30
fix(Vectorizer): handle implicit coordinates when converting lines to path data

bendemboski:
## Description

If a line is missing any of its start/end coordinate attributes, it defaults to 0, but the `convertLineToPathData()` wasn't filling in zeros, but instead leaving missing attributes as the empty string, resulting in corrupt path data. So, default missing values to `0`.


kumilingus:
Thank you, @bendemboski, as always. Appreciate your contribution.

PR #2997, opened 2025-06-17
chore(joint-react): fix IDs in jest snapshots

zbynekstara:


samuelgja:
https://github.com/clientIO/joint/pull/2998 this PR fix the issue with changing ids

PR #3013, opened 2025-07-01
fix(ports): parse numeric string args

MartinKanera:
## Description

Normalize `x` and `y` port-position arguments given as plain numeric strings (e.g. `{ x: '107', y: '15' }`) by converting them to numbers (e.g. `{ x: 107, y: 15 }`).

## Motivation and Context

Previously, calling `link.getBBox()` could produce incorrect values because port-position calculations relied on `Point.offset(dx, dy)`, which performs addition on the coordinates. When the current `x`/`y` coordinates of the `Point` instance were strings, this led to unintended string concatenation instead of numeric addition.

### DIFF
--- a/packages/joint-core/src/layout/ports/port.mjs
+++ b/packages/joint-core/src/layout/ports/port.mjs
@@ -1,12 +1,21 @@
 import * as g from '../../g/index.mjs';
 import * as util from '../../util/index.mjs';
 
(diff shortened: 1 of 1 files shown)


kumilingus (changes requested):
> Consider extracting the string-to-number conversion for x and y into a shared helper function to reduce duplicated code and improve readability.

PR #3025, opened 2025-07-30
feat(dia.Graph, dia.Paper): cell layers

Geliogabalus:
## Description

The PR adds cell layers functionality both to `dia.Graph` and `dia.Paper` classes. 

### feat(dia.Graph):

Add multiple API methods to support cell layers functionality. Cell layers are special cell groups with its own z-index context. In previous version there was only one cell layer posible - now it is represented by default cell layer with id `cells`. All cell layers are instances of `dia.CellLayer` class which is a child of a generic `dia.CellGroup` model. One of the cell layers is designated as `default cell layer` - you can set it using `setDefaultCellLayer()` method. When you have cell without layer specified it will be added to the `default cell layer`.

All cell layers added to the Graph are rendered in the Paper as special views in the order specified in the cell layers collection. When using `addCellLayer()` method without options the cell layer is added to the end of the cell layer list. You can use `addCellLayer(layer, { insertBefore: 'layerId' })` to add/move layer to other position.

Besides adding layer as an `dia.CellLayer` instance you can add it as JSON object similar to how cells are processed. To inform Graph which types to use when constructing new instances, `cellLayerNamespace` graph option is used. By default, `dia.CellLayer` type is used. 

```ts
graph.addCellLayer({
    id: 'layer_1'
})
```

#### events

To listen for cell layers changes Graph emits events with the `layers:` prefix. These events are propagated model events from each `dia.CellLayer` and the whole `cellLayersCollection`. For example, when the collection has been sorted (the order of the cell layers has changed) the `layers:sort` event is emitted.

### feat(dia.Cell):

Add `layer()` method which controls the layer of the particular cell. By default, the layer id is stored in the `layer` attribute, but can be changed using `config.layerAttribute` option. When the attribute is not set or set to `null` the `layer()` returns default cell layer id if the cell is added to the Graph.

Now `toFront()` and `toBack()` functions move the cell within its layer.

### feat(dia.Paper):

Added multiple methods for manipulating layer views. Every layer view is an instance of `dia.LayerView` class. 

You can insert new layer views into the Paper using `insertLayerView()` method. When using `insertLayerView()` method without options the layer view is added to the end of the `layers` element. You can use `insertLayerView(layer, { insertBefore: 'layerId' })` to insert/move layer view to other positions.

New cell layers are created in Paper automatically. For every `dia.CellLayer` added to the Graph corresponded `dia.CellLayerView` is created. All cell layer views are placed between `back` and `labels` layer views.

You can create new layer view using `createLayerView()` method. This method creates layer view instance using provided options. It will also inject `paper` reference to the layer view. To inform Paper which types to use when constructing new instances, `layerViewNamespace` paper option is used. By default, `dia.LayerView` type is used. 

```ts
const layerView = paper.createLayerView({
    id: 'layer_view_1',
})
```

When using View with cell layer model, the type will be constructed from cell layer type similar to how `dia.CellView` instances are created. For example, if cell layer has type `CustomCellLayer`, Paper will look for `CustomCellLayerView` type in `layerViewNamespace`.

### feat(dia.CellGroup, dia.CellLayer): 

New classes representing cell collection has been added.

### feat(dia.LayerView, dia.CellLayerView, dia.GridLayerView):

New classes representing layer views have been added. 

__LayerView and GridLayerView have been in JointJS before, but were not exposed.

### feat(dia.GraphCells, dia.GraphCellLayers):

These collection properly exposed to users. You can get instances of these collections using `cellsCollection` and `cellLayersCollection` graph properties.

### DIFF
--- a/packages/joint-core/src/dia/Paper.mjs
+++ b/packages/joint-core/src/dia/Paper.mjs
@@ -2102,6 +2102,7 @@ export const Paper = View.extend({
         }
 
         // combine all content area rectangles from all cell layers
+        // using only cell layer views to avoid including non-cell view content (e.g. grid)
         const bbox = g.Rect.fromRectUnion(...cellLayerViews.map(cellLayerView => V(cellLayerView.el).getBBox()));
         return bbox;
     },
--- a/packages/joint-core/src/dia/collections/GraphCells.mjs
+++ b/packages/joint-core/src/dia/collections/GraphCells.mjs
@@ -22,6 +22,8 @@ export const GraphCells = Collection.extend({
         this.graph = opt.graph;
     },
 
+    // Overriding the default `model` method to create cell models
+    // based on their `type` attribute and the `cellNamespace` option.
     model: function(attrs, opt) {
 
         const collection = opt.collection;
@@ -37,6 +39,8 @@ export const GraphCells = Collection.extend({
         return new ModelClass(attrs, opt);
     },
 
+    // Override Collection's _addReference
+    // to set the graph reference on the model.
     _addReference: function(model, options) {
         Collection.prototype._addReference.apply(this, arguments);
         // If not in `dry` mode and the model does not have a graph reference yet,
@@ -46,6 +50,8 @@ export const GraphCells = Collection.extend({
         }
     },
 
+    // Override Collection's _removeReference
+    // to remove the graph reference from the model.
     _removeReference: function(model, options) {
         Collection.prototype._removeReference.apply(this, arguments);
         // If not in `dry` mode and the model has a reference to this exact graph,
@@ -55,8 +61,9 @@ export const GraphCells = Collection.extend({
         }
     },
 
-    // fast version for reset function
-    // it does not clear the `_byId` map
+    // fast version of _removeReference for reset method.
+    // It does not clear the `_byId` map because it is cleared as part of the reset.
+    // TODO: move "fast" logic to the `mvc.Collection` in later versions.
     _removeReferenceFast: function(model, options) {
         if (this === model.collection) delete model.collection;
         model.off('all', this._onModelEvent, this);
--- a/packages/joint-core/src/dia/controllers/CellLayersController.mjs
+++ b/packages/joint-core/src/dia/controllers/CellLayersController.mjs
@@ -20,12 +20,11 @@ export class CellLayersController extends Listener {
         this.collection = this.graph.cellLayerCollection;
 
         // Default setup
-        this.defaultCellLayerId = DEFAULT_CELL_LAYER_ID;
-        this.graph.trigger('layers:default:change', this.graph, this.defaultCellLayerId);
         this.addCellLayer({
             id: DEFAULT_CELL_LAYER_ID,
             __legacy: true
         });
+        this.defaultCellLayerId = DEFAULT_CELL_LAYER_ID;
 
         this.startListening();
     }
--- a/packages/joint-core/src/mvc/Model.mjs
+++ b/packages/joint-core/src/mvc/Model.mjs
@@ -26,10 +26,7 @@ export var Model = function(attributes, options) {
     var attrs = attributes || {};
     options || (options = {});
 
-    this.eventPrefix = '';
-    if (options.eventPrefix) {
-        this.eventPrefix = options.eventPrefix;
-    }
+    this.eventPrefix = options.eventPrefix || '';
 
     this.preinitialize.apply(this, arguments);
     this.cid = uniqueId(this.cidPrefix);
--- a/packages/joint-core/types/joint.d.ts
+++ b/packages/joint-core/types/joint.d.ts
@@ -207,11 +207,11 @@ export namespace dia {
             [key: string]: any;
         }
 
-        type ResetCellLayersOptions = Options & {
+        interface ResetCellLayersOptions extends Options {
             defaultCellLayer?: string;
         }
 
-        type AddCellLayerOptions = Options & {
+        interface AddCellLayerOptions extends Options {
             insertBefore?: string;
         }
     }


packages/joint-core/src/dia/ElementView.mjs
```diff
@@ -364,7 +364,6 @@ export const ElementView = CellView.extend({
 
         element.startBatch('to-front');
 
-        // Bring the model to the front with all his embeds.
```
kumilingus, line 367: missing comment

packages/joint-core/src/dia/Cell.mjs
```diff
@@ -956,6 +960,31 @@ export const Cell = Model.extend({
...
+            if (layerId == null && this.graph) {
+                layerId = this.graph.getDefaultCellLayer().id;
+            }
+
+            return layerId;
+        }
+
+        // otherwise set the layer id
+        if (!isString(layerId)) {
+            throw new Error('Layer id must be a string.');
```
kumilingus: `dia.Cell: Layer id must be a string.`

packages/joint-core/src/dia/Graph.mjs
```diff
@@ -429,35 +423,83 @@ export const Graph = Model.extend({
...
     getCell: function(id) {
 
-        return this.get('cells').get(id);
+        return this.cellCollection.get(id);
     },
 
     getCells: function() {
-
-        return this.get('cells').toArray();
+        // Preserve old order without layers
```
kumilingus: What does `old order` mean? The reader would not know.

packages/joint-core/src/dia/Graph.mjs
```diff
@@ -429,35 +423,83 @@ export const Graph = Model.extend({
...
 
-        return this.get('cells').toArray().filter(cell => cell.isLink());
+        return this.getCells().filter(cell => cell.isLink());
     },
 
-    getFirstCell: function() {
+    getFirstCell: function(layerId) {
+        let cells;
+        if (!layerId) {
+            cells = this.getCells();
```
kumilingus: Can we do this more efficiently? 
```ts
this.cellLayersController.getCellLayer(firstLayerId).cells;
```

packages/joint-core/src/dia/Graph.mjs
```diff
@@ -70,23 +64,20 @@ export const Graph = Model.extend({
 
         opt = opt || {};
 
+        this.cellLayersController = new CellLayersController({ graph: this });
+
         // Passing `cellModel` function in the options object to graph allows for
         // setting models based on attribute objects. This is especially handy
         // when processing JSON graphs that are in a different than JointJS format.
-        var cells = new GraphCells([], {
+        this.cellCollection = new GraphCells([], {
```
kumilingus: Consider adding a `cellCollection` variable:
```ts
const cellCollection = this.cellCollection = new GraphCells([],{});`
```

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -48,9 +48,17 @@ import * as connectionPoints from '../connectionPoints/index.mjs';
 import * as anchors from '../anchors/index.mjs';
 
 import $ from '../mvc/Dom/index.mjs';
-import { GridLayer } from './layers/GridLayer.mjs';
+import { GridLayerView } from './layers/GridLayerView.mjs';
+
+export const LAYERS = {
```
kumilingus: Do we name enums with all upper-case letter anywhere?

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -437,7 +451,7 @@ export const Paper = View.extend({
         };
 
         // Render existing cells in the graph
-        this.resetViews(model.attributes.cells.models);
+        this.resetViews(model.cellCollection.models);
```
kumilingus: Is this a breaking change? `cellCollection.models` is not sorted, right?

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -526,10 +522,34 @@ export const Paper = View.extend({
...
+                layerView = this.renderLayerView({
+                    id: cellLayer.id,
+                    model: cellLayerModel
+                });
+
+                this._cellLayerViews[cellLayer.id] = layerView;
+            } else {
+                layerView = this.getLayerView(cellLayer.id);
+            }
+            this.insertLayerView(layerView, LAYERS.LABELS);
```
kumilingus: Why `Layers.LABELS`? Can you add a comment to the code here.

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -526,10 +522,34 @@ export const Paper = View.extend({
...
             }
         }
-        var sortDelayingBatches = this.SORT_DELAYING_BATCHES;
-        if (sortDelayingBatches.includes(name) && !graph.hasActiveBatch(sortDelayingBatches)) {
-            this.sortViews();
-        }
+    },
+
+    updateCellLayers: function(cellLayers) {
+        const removedCellLayerViewIds = this._cellLayers.filter(cellLayerView => !cellLayers.some(l => l.id === cellLayerView.id)).map(cellLayerView => cellLayerView.id);
```
kumilingus: line is too long

kumilingus: This approach to removing leftovers isn’t the most optimal, but given the expected number of layers, it should be sufficient. Please add a comment to the code to indicate that while this can be improved, it’s acceptable for our current use case.

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -616,121 +636,136 @@ export const Paper = View.extend({
...
-    addLayer(layerName, layerView, options = {}) {
-        if (!layerName || typeof layerName !== 'string') {
-            throw new Error('dia.Paper: The layer name must be provided.');
-        }
-        if (this._getLayerView(layerName)) {
-            throw new Error(`dia.Paper: The layer "${layerName}" already exists.`);
+    requestLayerViewRemove(layerView) {
+        layerView = this._requireLayerView(layerView);
+
+        // TODO: change when after view management merge
```
kumilingus: TODO

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -616,121 +636,136 @@ export const Paper = View.extend({
...
-        // Returns a sorted array of layer names.
+    getLayerViewOrder() {
+        // Returns a sorted array of layer view ids.
         return this._layers.order.slice();
     },
 
-    getLayers() {
-        // Returns a sorted array of layer views.
-        return this.getLayerNames().map(name => this.getLayerView(name));
+    getOrderedLayerViews() {
```
kumilingus: Do we have `getLayerViews()` method that is unordered?

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -616,121 +636,136 @@ export const Paper = View.extend({
...
+        if (!this.hasLayerView(layerId)) {
+            throw new Error(`dia.Paper: Unknown layer view "${layerId}".`);
         }
-        const { insertBefore } = options;
+
+        const { _layers: { order }} = this;
+
         if (!insertBefore) {
-            this._registerLayer(layerName, layerView, null);
+            // remove from order in case of a move command
```
kumilingus: Can you explain what is the `move command`?
It will be better to refer to it as `in case we are changing position of an existing layer`

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -768,46 +803,68 @@ export const Paper = View.extend({
...
+    removeLayerViews: function() {
         const { _layers: { viewsMap }} = this;
-        Object.values(viewsMap).forEach(layerView => this._removeLayer(layerView));
+        Object.values(viewsMap).forEach(layerView => this._removeLayerView(layerView));
     },
 
-    resetLayers: function() {
+    resetLayerViews: function() {
         const { _layers: { viewsMap }} = this;
         Object.values(viewsMap).forEach(layerView => layerView.removePivots());
```
kumilingus: I think it would be nice to have `reset()` method on LayerView (which will call `removePivots()` internally)

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -1520,7 +1583,7 @@ export const Paper = View.extend({
         }
         this.options.frozen = updates.keyFrozen = false;
         if (updates.sort) {
-            this.sortViews();
+            this.sortLayers();
```
kumilingus: Should we call this `sortLayerViews()`?

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -1791,7 +1854,16 @@ export const Paper = View.extend({
             return this.model.getBBox() || new Rect();
         }
 
-        return V(this.cells).getBBox();
+        const cellLayerViews = Object.values(this._cellLayerViews);
```
kumilingus, line 2097: Can we use `Rect.fromRectUnion()` for simplicity here?
Also, please add a comment what is being done here.

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -2029,41 +2100,25 @@ export const Paper = View.extend({
             this._updates.sort = true;
             return;
         }
-        this.sortViewsExact();
+        this.sortLayersExact();
```
kumilingus: Same here: `sortLayerViewsExact()`

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -2029,41 +2100,25 @@ export const Paper = View.extend({
...
+        const { _layers: { viewsMap }} = this;
 
-        sortElements(cellNodes, function(a, b) {
-            var cellA = cells.get(a.getAttribute('model-id'));
-            var cellB = cells.get(b.getAttribute('model-id'));
-            var zA = cellA.attributes.z || 0;
-            var zB = cellB.attributes.z || 0;
-            return (zA === zB) ? 0 : (zA < zB) ? -1 : 1;
-        });
+        Object.values(viewsMap).filter(view => view instanceof CellLayerView).forEach(view => view.sortExact());
```
kumilingus: Don't we have a higher-level API to get all CellLayerViews?

packages/joint-core/src/dia/Paper.mjs
```diff
@@ -2029,41 +2100,25 @@ export const Paper = View.extend({
...
-                layerView.insertSortedNode(el, model.get('z'));
-                break;
-            case sortingTypes.EXACT:
-            default:
-                layerView.insertNode(el);
-                break;
-        }
+        layerView.insertCellView(view);
+
+        this.trigger('cell:inserted', view, isInitialInsert);
```
kumilingus: Perhaps we should call it `'cell:insert'` (we don't use `added`, `changed` either).

Or better `'cell:mount'`. Analogous to the `onMount()` method.

kumilingus: Or remove it completely, if it is not used in this PR.

packages/joint-core/src/dia/groups/CellGroup.mjs
```diff
@@ -0,0 +1,67 @@
...
+        this.cells.remove(cell, opt);
+    }
+
+    reset() {
+        this.cells.toArray().forEach(cell => {
+            this.remove(cell);
+        });
+    }
+
+    groupSet(key, val, opt) {
```
kumilingus: Please remove the method if it is not used. Let's not commit to an API we are not sure about yet.

packages/joint-core/src/dia/groups/CellGroup.mjs
```diff
@@ -0,0 +1,67 @@
...
+
+    add(cell, opt) {
+        this.cells.add(cell, opt);
+    }
+
+    remove(cell, opt) {
+        this.cells.remove(cell, opt);
+    }
+
+    reset() {
```
kumilingus: The method removes cell by cell and does not allow replacing cells with a new set.
A better name could be `clear()`. Similarly to `graph.clear()`.

packages/joint-core/src/dia/groups/CellGroup.mjs
```diff
@@ -0,0 +1,67 @@
...
+            // do not change collection attribute of a cell model
+            return attrs;
+        } else {
+            throw new Error('CellGroupCollection only accepts Cell instances.');
+        }
+    }
+
+    _onModelEvent(event, model, collection, options) {
+        if (model) {
+            if (event === 'changeId') {
```
kumilingus: You can refactor `Collection.mjs` if you need to partially  preserved its behavious.

Why do we omit this line?
```ts
if ((event === 'add' || event === 'remove') && collection !== this) return;
```

packages/joint-core/src/dia/groups/CellGroup.mjs
```diff
@@ -0,0 +1,67 @@
+import { Model, Collection } from '../../mvc/index.mjs';
+
+export class CellGroupCollection extends Collection {
+
+    _prepareModel(attrs, _options) {
+        if (this._isModel(attrs)) {
+            // do not change collection attribute of a cell model
```
kumilingus: Why did you omit this line?
```ts
if (!attrs.collection) attrs.collection = this;
```

These are low-level features and changing them must be properly explained in the code.

packages/joint-core/src/dia/groups/CellLayer.mjs
```diff
@@ -0,0 +1,37 @@
...
+        return model.get('z') || 0;
+    }
+}
+
+export class CellLayer extends CellGroup {
+
+    defaults() {
+        return {
+            type: 'CellLayer',
+            collectionConstructor: CellLayerCollection,
```
kumilingus: Do we need to store a JS class inside the model attributes (`CellGroup` is a model)?

packages/joint-core/src/dia/layers/CellLayerView.mjs
```diff
@@ -0,0 +1,111 @@
...
+import { sortingTypes } from '../Paper.mjs';
+
+export const CellLayerView = LayerView.extend({
+
+    SORT_DELAYING_BATCHES: ['add', 'reset', 'to-front', 'to-back'],
+
+    init() {
+        LayerView.prototype.init.apply(this, arguments);
+
+        // applying it here instead of using property
```
kumilingus: Can we update the snapshot instead?

packages/joint-core/src/dia/layers/CellLayerView.mjs
```diff
@@ -0,0 +1,111 @@
...
+                this.insertSortedNode(el, model.get('z'));
+                break;
+            case sortingTypes.EXACT:
+            default:
+                this.insertNode(el);
+                break;
+        }
+    },
+
+    _requestViewUpdate(cell, opt) {
```
kumilingus: This is not just an update request. This is a request for inserting / re-inserting.
This should be called `_requestCellViewInsertion()`.

packages/joint-core/src/dia/Graph.mjs
Geliogabalus: Did comment it properly

packages/joint-core/src/dia/Paper.mjs
Geliogabalus: Yes, indeed. Changed to `model.getCells()`, this one returns ordered models

packages/joint-core/src/dia/Paper.mjs
Geliogabalus: Added the comment. I am putting cell layers between `back` and `labels` layers, where `cells` layer is currently located

packages/joint-core/src/dia/Paper.mjs
Geliogabalus: Added `getLayerViews()` and `getCellLayerViews` methods.

packages/joint-core/src/dia/Paper.mjs
Geliogabalus: Added `getCellLayerViews()` method

packages/joint-core/src/dia/Paper.mjs
Geliogabalus: Removed in this PR

packages/joint-core/src/dia/groups/CellGroup.mjs
Geliogabalus: This method is used when I need to change `layer` attribute for all cells, when the default layer has changed

packages/joint-core/src/dia/groups/CellGroup.mjs
Geliogabalus: There is a `clear` method on mvc.Model, so I resorted to this naming

packages/joint-core/src/dia/groups/CellGroup.mjs
Geliogabalus: Added a comment. It is done, so the CellGroup can emit `add` and `remove` events, because the `collection` is normally `graph.cellCollection` origin collection

packages/joint-core/src/dia/groups/CellLayer.mjs
Geliogabalus: I removed this attribute and refactored cellLayers attribute in a way that it stores all layer attributes similar to cells.

packages/joint-core/test/jointjs/layers/basic.js
```diff
@@ -94,6 +94,8 @@ QUnit.module('layers-basic', function(hooks) {
             ]
         });
 
+        console.log(joint.config)
```
kumilingus: leftover

PR #3082, opened 2025-10-22
feat(dia.Graph): remove base cell collection

Geliogabalus:
## Description

Remove base cell collection and move logic to the cell layers instead

### events

All cell events are emitted without prefix.
All layer model events and cell collection events are emitted with `layer:` prefix, e.g. `layer:reset` is emited when layer collection has been reset, and `layer:change` when the layer's attribute has been changed.
All layer collection events are emited with `layers:` prefix, e.g. `layers:reset` when whole cellLayerCollection has been reset.

### DIFF
--- a/packages/joint-core/src/dia/collections/CellGroupCollection.mjs
+++ b/packages/joint-core/src/dia/collections/CellGroupCollection.mjs
@@ -20,8 +20,8 @@ export class CellGroupCollection extends Collection {
     }
 
     modelId(attrs, idAttribute) {
-	    return attrs[idAttribute || 'id'];
-	}
+        return attrs[idAttribute || 'id'];
+    }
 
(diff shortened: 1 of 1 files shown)


packages/joint-core/src/dia/Graph.mjs
```diff
@@ -202,30 +228,28 @@ export const Graph = Model.extend({
...
 
-        if (collection.length === 0) return this;
+        if (cells.length === 0) return this;
 
         this.startBatch('clear', opt);
 
-        // The elements come after the links.
-        var cells = collection.sortBy(function(cell) {
-            return cell.isLink() ? 1 : 2;
+        const sortedCells = cells.sort((a, b) => {
```
kumilingus: use `util.sortBy`

PR #3093, opened 2025-10-30
fix(theme): correct highlight color to match design specifications

hartou:
fix(layout-msagl): update nested option label for clarity

fix(presets): rename nested graph title to Microservice Architecture

refactor(list): change itemPosition return type for better type safety

fix(arrow): correct SVG import filename to match casing

fix(plus): correct SVG import filename to match casing

## Description

Clearly and concisely describe your changes in detail.

## Motivation and Context

Why is this change required? What problem does it solve? If applicable, describe the new features the solution brings.


kumilingus:
Hi @hartou , thanks for the PR! Could you share the motivation behind these changes? At first glance, it looks like it may have been generated automatically, so any context would be helpful.

PR #3100, opened 2025-11-04
fix(dia.LayerView): ensure LayerView lifecycle methods receive correct arguments

kumilingus:
## Description

`LayerView` lifecycle methods (`afterPaperReferenceSet` and `beforePaperReferenceUnset`) now execute with expected arguments.


packages/joint-core/types/joint.d.ts
```diff
@@ -2333,7 +2333,7 @@ export namespace dia {
 
         protected afterPaperReferenceSet(paper: Paper): void;
 
-        protected beforePaperReferenceUnset(paper: Paper): void;
```
samuelgja, line 2336: you removed beforePaperReferenceUnset, but we are using https://github.com/kumilingus/joint/blob/567196e3b88b885fe83de37a4a91e0131750182d/packages/joint-core/src/dia/GridLayerView.mjs#L32

So its not better to use
```ts
beforePaperReferenceUnset(paper?: Paper): void
```
?

kumilingus: The paper is now always sent.

PR #3117, opened 2025-12-10
fix(layout-directed-graph): ignore cell layers in the layout

kumilingus:
## Description

The current layout throws an error when a cell uses a custom layer, because that layer doesn’t exist on the temporary layout graph created internally.

This PR adds the `ignoreLayers` graph option, which causes the graph to ignore cell layers and place all cells in the default layer. This is useful in scenarios where layers have no significance (e.g., automatic layouts), making it unnecessary to create them just to add cells.

```ts
const graph = new dia.Graph(graphAttributes, { ignoreLayers: true });
```

### DIFF
--- a/packages/joint-core/test/jointjs/graph.js
+++ b/packages/joint-core/test/jointjs/graph.js
@@ -2578,20 +2578,51 @@ QUnit.module('graph', function(hooks) {
         );
     });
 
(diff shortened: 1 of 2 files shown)


Geliogabalus (approved):
Look good

PR #3120, opened 2025-12-15
fix(layout-msagl): ignore cell layers in the layout

kumilingus:
## Description

The current layout throws an error when a cell uses a custom layer, because that layer doesn’t exist on the temporary layout graph created internally.

This PR adds the `ignoreLayers` graph option to the internally created graph, preventing issues related to missing layers.


MartinKanera (approved):
Looks good to me.

PR #3150, opened 2026-01-27
example(joint-react): add control element tool example

MartinKanera:


packages/joint-react/src/stories/examples/with-element-controls/code.tsx
```diff
@@ -0,0 +1,214 @@
...
+function Parallelogram({ width, height, offset = 0, label }: Readonly<ElementData>) {
+  const parallelogramSvgPath = useMemo(() => {
+    return `
+      M 0 ${height}
+      L ${offset} 0
+      L ${width} 0
+      L ${width - offset} ${height}
+      Z
+    `;
+  }, [width, height, offset]);
```
samuelgja: This can be also inline without memo, its cheap i guess. Memo should be used just with heavy stuff like find something in array or work with collections - just to let you know, but its okey.

packages/joint-react/src/stories/examples/with-element-controls/code.tsx
```diff
@@ -0,0 +1,214 @@
...
+      v -${height / 2 - thickness / 2}
+      H 0
+      v -${thickness}
+      H ${width - arrowHeight}
+      z
+    `;
+  }, [width, height, arrowHeight, thickness]);
+
+  return (
+    <g>
```
samuelgja: i think its good, but renderElement is already inside G, so if you inspect html,  I guess this provides g inside g, so maybe use fragment (<>)

packages/joint-react/src/stories/examples/with-element-controls/code.tsx
```diff
@@ -0,0 +1,214 @@
...
+      renderElement={renderElement}
+      onElementsSizeReady={addElementControls}
+    />
+  );
+}
+
+export default function App() {
+  return (
+    <GraphProvider
+      areBatchUpdatesDisabled={true}
```
samuelgja: use just areBatchUpdatesDisabled, without areBatchUpdatesDisabled={true} in react thats already boolean true.

packages/joint-react/src/stories/examples/with-element-controls/code.tsx
MartinKanera: Yeah, you are right

packages/joint-react/src/stories/examples/with-element-controls/story.tsx
```diff
@@ -0,0 +1,18 @@
...
+export type Story = StoryObj<typeof Code>;
+
+export default {
+  title: 'Examples/Element Controls',
+  component: Code,
+  tags: ['example'],
+  parameters: makeRootDocumentation({
+    code: CodeRaw,
+  }),
+} satisfies Meta<typeof Code>;
```
samuelgja, line 16: add explanation either here via:
sometjing like:
 parameters: makeRootDocumentation({
    description: `
Custom is a component that allows you to use a custom highlighter. You must provide the \`onAdd\` which must return jointjs highlighter.
    `,
    apiURL: API_URL,
    code: `import { Highlighter } from '@joint/react'
<Highlighter.Custom
  onAdd={...}
  options={{ ... }}
>
  <rect rx={10} ry={10} width={100} height={50} fill={"blue"} />
</Highlighter.Custom>
    `,
  }),

Or create docs.mdx file for this story - just little docs about what is it doing,

PR #3151, opened 2026-01-29
fix(dia.Paper): handle element removal during pointer events

Geliogabalus:
## Summary
- Fixes an issue where removing an element during `pointermove` or `pointerup` event handlers would cause errors
- When `findViewByModel()` returns null (because the model was removed), falls back to using the original view reference

## Problem
When an element is removed from the graph during a pointer event handler (e.g., deleting an element after dragging), subsequent calls to `view.pointermove()` or `view.pointerup()` would fail because `findViewByModel(view.model)` returns `undefined` for removed models.

## Solution
Added a fallback to use the original view reference when `findViewByModel()` returns falsy:

```javascript
// Before
view = this.findViewByModel(view.model);

// After
view = this.findViewByModel(view.model) || view;
```

This change is applied to both `pointermove` and `pointerup` handlers in `packages/joint-core/src/dia/Paper.mjs`.

### DIFF
--- a/packages/joint-core/test/jointjs/paper.js
+++ b/packages/joint-core/test/jointjs/paper.js
@@ -2974,9 +2974,9 @@ QUnit.module('paper', function(hooks) {
             const testElementView = testElement.findView(paper);
             const testElementRect = testElementView.el.querySelector('rect');
 
-            // Second handler: Should still be called even though element was removed
             paper.once('element:pointerup', spy);
 
+            // Override pointerup to remove the element during the event
             paper.pointerup = function(evt) {
                 testElement.remove();
                 joint.dia.Paper.prototype.pointerup.call(paper, evt);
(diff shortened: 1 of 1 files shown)


packages/joint-core/test/jointjs/paper.js
```diff
@@ -2959,6 +2959,90 @@ QUnit.module('paper', function(hooks) {
...
+            const testElement = new joint.shapes.standard.Rectangle({
+                position: { x: 100, y: 100 },
+                size: { width: 100, height: 100 }
+            });
+            testElement.addTo(graph);
+
+            const testElementView = testElement.findView(paper);
+            const testElementRect = testElementView.el.querySelector('rect');
+
+            document.addEventListener('mousemove', () => {
```
kumilingus: Can this be done instead?

```ts
simulate.mousedown({ el: testElementRect });
simulate.mousemove({ el: testElementRect });
testElement.remove();
simulate.mouseup({ el: testElementRect });
```

Geliogabalus: I checked, and with this setup it never even goes into `paper.pointerup` function when event is triggered, so this is meaningless I think.

Geliogabalus: I think it is important that at the start of the event chain, the view is in the dom as a part of the paper.

PR #3159, opened 2026-01-30
fix(joint-react): fix links flickering

samuelgja:
## Description

Fixes link flickering issue where links would briefly appear before element content was rendered.

## Motivation and Context

**The Bug:** When rendering a Paper with elements and links, the visual sequence was:
1. Frame 1: Empty screen
2. Frame 2: Only links visible (pointing to positions, but no element content)
3. Frame 3: Correct render with both elements and links

This happened because JointJS renders link SVG paths synchronously, while React portals render element content via microtask (later).

### Changes

- Refactored `create-elements-size-observer.ts` to ensure consistent timing between link paths and element content
- Added comprehensive test suite (`paper-link-flickering.test.tsx`) that verifies the invariant: "If links are visible with rendered paths, elements must also have their content rendered"
- Added unit tests for the size observer

### DIFF
--- a/packages/joint-react/CLAUDE.md
+++ b/packages/joint-react/CLAUDE.md
@@ -1,0 +1,103 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+`@joint/react` is the React bindings library for JointJS. It provides React components and hooks for building diagramming applications with JointJS, rendering diagram elements as React components via SVG foreignObject portals.
+
+**Stack:** TypeScript 5.9, React 19 (peer: >=18 <20), esbuild (build), Vite (storybook/dev), Jest 30 + @testing-library/react (testing), ESLint 9 flat config, Storybook 10
+
+## Common Commands
+
+```bash
+# Build (esbuild → dist/cjs, dist/esm, dist/types)
+yarn build
+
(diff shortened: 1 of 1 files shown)


packages/joint-react/src/components/paper/paper.types.ts
```diff
@@ -13,7 +13,7 @@ export interface OnLoadOptions {
 
 type ReactPaperOptionsBase = OmitWithoutIndexSignature<
   dia.Paper.Options,
-  'frozen' | 'defaultLink' | 'autoFreeze' | 'viewManagement'
+  'frozen' | 'defaultLink' | 'autoFreeze' | 'viewManagement' | 'measureNode'
```
samuelgja, line 16: I am not sure about this @kumilingus

packages/joint-react/src/store/__tests__/create-elements-size-observer.test.ts
```diff
@@ -0,0 +1,288 @@
+/* eslint-disable prefer-destructuring */
```
samuelgja, line 1: Add immediate handling when an element is added to the observer.
The observer is attached instantly, but it only detects changes on the next animation frame.
To eliminate this delay, we immediately measure the HTML element, while still processing measurements in a batch.

kumilingus (approved):

PR #3266, opened 2026-04-07
feat(router): add minPathMargin option to rightAngle router to suppress margin-overlap detours

Geliogabalus:
## Summary

- Add a `minPathMargin` option to `rightAngleRouter`
- Add `sourceMargin` and `targetMargin` options to `rightAngleRouter`
- Refactor codebase to increase readablity.
- Add a `right-angle-playground-js` example demonstrating the option with movable vertices, anchors, and a resize handle

## Test plan

- [ ] Verify that omitting `minPathMargin` (default `null`) preserves existing routing behaviour in all side combinations
- [ ] Verify new `sourceMargin` and `targetMargin` options
- [ ] Verify the `right-angle-playground-js` example loads and the router respects the `minPathMargin: 10` setting when elements are dragged close together


packages/joint-core/src/routers/rightAngle.mjs
```diff
@@ -543,9 +548,16 @@ function routeBetweenPoints(source, target, opt = {}) {
 
             const isUpwardsShorter = topD < bottomD;
 
+            let ignoreMargin = false;
+            if (ignoreOverlappingMargin &&
```
MartinKanera, line 552: Is it intentional to skip the `ignoreMargin` path when `ignoreOverlappingMargin === 0`?

packages/joint-core/types/joint.d.ts
```diff
@@ -4152,6 +4152,7 @@ export namespace routers {
 
     interface RightAngleRouterArguments {
         margin?: number;
+        minMargin?: number;
```
MartinKanera: We are setting the minMargin to `null`, shouldn't we allow the explicit null as well here?

packages/joint-core/src/routers/rightAngle.mjs
Geliogabalus: Yes, in the case of `ignoreOverlappingMargin === 0` this is the same check as the original margin check without this option

packages/joint-core/types/joint.d.ts
Geliogabalus: Added `null` to the typings.

kumilingus (dismissed):

kumilingus (approved):

PR #3297, opened 2026-04-28
feat(joint-core): predicate form for Cell.toJSON ignoreEmptyAttributes

kumilingus:
## Summary
- `Cell.toJSON({ ignoreEmptyAttributes })` now also accepts a predicate `(key: string, path: string[]) => boolean` for fine-grained control over which empty `{}` values are pruned from the serialized output.
- `true` keeps its existing top-level-only semantics — sugar for `(key, path) => path.length === 1`. No breaking change.
- Predicate is invoked bottom-up; a parent emptied by child removal is itself a candidate, so nested empties (e.g. `{ ports: { groups: {} } }`) are reachable when the predicate opts in.

## Motivation
The boolean form was all-or-nothing and only stripped top-level empties. There was no way to:
- Drop nested empties (`{ ports: { groups: {} } }` survived).
- Keep a known set of empties (e.g. an intentional `attrs.text.textWrap: {}`) while pruning the rest.

## Example
```js
cell.toJSON({
  ignoreDefaults: true,
  // keep empties at depth 3 inside `attrs`, drop everything else empty
  ignoreEmptyAttributes: (key, path) => !(path[0] === 'attrs' && path.length === 3),
});
```

## Test plan
- [x] New QUnit test in `test/jointjs/cell.js` covering the predicate form.
- [x] Existing `ignoreEmptyAttributes: true` tests still pass (1946 / 1946).
- [x] Type definition updated in `dia.d.ts` (`Cell.ExportOptions`).

### DIFF
--- a/packages/joint-core/types/dia.d.ts
+++ b/packages/joint-core/types/dia.d.ts
@@ -614,7 +614,7 @@ export namespace Cell {
 
     interface ExportOptions {
         ignoreDefaults?: boolean | string[];
-        ignoreEmptyAttributes?: boolean;
+        ignoreEmptyAttributes?: boolean | ((key: string, path: string[]) => boolean);
     }
 
     type UnsetCallback<V> = (
--- a/packages/joint-core/types/dia.d.ts
+++ b/packages/joint-core/types/dia.d.ts
@@ -612,9 +612,11 @@ export namespace Cell {
         mergeArrays?: boolean;
     }
 
+    type IgnoreEmptyAttributesCallback = (key: string, path: string[]) => boolean;
+
     interface ExportOptions {
         ignoreDefaults?: boolean | string[];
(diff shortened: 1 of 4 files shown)


packages/joint-core/types/dia.d.ts
```diff
@@ -614,7 +614,7 @@ export namespace Cell {
 
     interface ExportOptions {
         ignoreDefaults?: boolean | string[];
-        ignoreEmptyAttributes?: boolean;
+        ignoreEmptyAttributes?: boolean | ((key: string, path: string[]) => boolean);
```
MartinKanera: I would extract the type here, something like:

```ts
type IgnoreEmptyAttributesPredicate = (key: string, path: string[]) => boolean;
```

MartinKanera (changes requested):

MartinKanera (approved):

PR #3306, opened 2026-05-05
fix(joint-react): register paper features synchronously during render

kumilingus:
## Summary

Fixes a class of bugs where features registered by hooks called inside a `<Paper>` subtree, or through a wrapping component (e.g. `<PaperScroller>` wrapping `<Paper>`), were not visible to sibling consumers in the same render pass. Adds adjacent fixes that surfaced while validating the main change.

### 1. Synchronous render-phase feature registration (`use-create-features.ts`)

Previously `useCreateFeature` deferred registration to `useLayoutEffect`. When a feature-registering component (e.g. `<Snaplines />`) mounted **inside** `<Paper>`, sibling consumers reading `usePaperFeatures(paperId)` in the same render captured an empty snapshot in their `onLoad` / imperative-API closures and never picked up the feature.

Fix: when `target === 'paper'` AND a `paperStore` is reachable AND the feature isn't already cached, register synchronously during render via `createAndRegisterFeature` + `registerFeature`. Mirrors the existing render-phase `featureContext.features.set(...)` path used when no `paperStore` exists. The `featureRef.current` guard in the existing useLayoutEffect re-register branch keeps `onAddFeature` firing exactly once.

### 2. Fallback paperStore lookup (`use-create-features.ts`)

When the hook is called **outside** a Paper subtree (e.g. `<PaperScroller>` wraps `<Paper>`, so the scroller sits above `PaperStoreContext`), `usePaperStore(OPTIONAL)` returns `null`. The deferred-queue path may have already registered the feature inside some `paperStore`, but `useCreateFeature` can't see it via context.

Fix: scan `graphStore.papers` for a paperStore whose `features` bag contains our `id`, fall back to it. Update / load paths now have a `paperStore` to operate on regardless of where the hook is mounted.

### 3. Existing-feature adoption (`use-create-features.ts`)

When the fallback paperStore resolves and a feature was already created via the deferred-queue path, the create-effect would have called `onAddFeature` again, producing duplicates. Adoption path: `resolveExistingFeature(...)` — if a feature with `id` already exists, store it in `featureRef.current` and skip creation.

### 4. `<Paper>` width/height routed through `setDimensions` (`use-create-portal-paper.tsx`)

`width` and `height` were spread into `assignOptions(paper.options, ...)`, which mutated `paper.options.width/height` directly and bypassed the DOM update. Worse, it tripped `setDimensions`'s early-exit guard the next time it ran (e.g. `<PaperScroller>` calling `paper.setDimensions(...)` after the prop change).

Fix: strip `width`/`height` out of the assignOptions spread and add a dedicated `useEffect` that calls `paper.setDimensions(width, height)`. Keeps DOM CSS and `paper.options` in lockstep.

### 5. `minPathMargin` option in `LinkRoutingOrthogonal` preset (`presets/link-routing.ts`)

Surfaces the new `minPathMargin` option from the `rightAngle` router (joint-core #3266) on the orthogonal preset typings.

### 6. New `@joint/react/stories` subpath export

Adds a stories utility entry point so `@joint/react-plus` story files can reuse `makeRootDocumentation`, `makeStory`, and `getAPILink` without reaching into `src/`. Not part of the runtime public API.

- `package.json` — `./stories` exports map entry
- `rollup.config.ts` — `src/stories.ts` added to entries
- `src/stories.ts` (new) — re-exports
- `src/internal.ts` — exposes `OPTIONAL` constant for downstream consumers

## Test plan

- [x] `yarn jest --testPathPatterns="features|use-create"` (43/43 pass)
- [x] `yarn typecheck` (no new errors)
- [ ] Manual: `<Snaplines />` mounted as child of `<Paper>` is detected by sibling `<Stencil />` (validated downstream in joint-plus)
- [ ] Manual: `<PaperScroller>` wrapping `<Paper>` — feature registration / update / unmount paths all hit the same `paperStore`
- [ ] Manual: changing `width`/`height` props on `<Paper>` resizes the DOM, and `<PaperScroller>` `setDimensions` calls afterwards still work

### DIFF
--- a/packages/joint-react/package.json
+++ b/packages/joint-react/package.json
@@ -36,6 +36,11 @@
       "import": "./dist/esm/presets/index.js",
       "require": "./dist/cjs/presets/index.js"
     },
+    "./stories": {
+      "types": "./dist/types/stories.d.ts",
+      "import": "./dist/esm/stories.js",
+      "require": "./dist/cjs/stories.js"
+    },
     "./src/*": "./src/*",
     "./scripts/*": "./scripts/*.ts",
     "./css/*": "./src/css/*"
--- a/packages/joint-react/rollup.config.ts
+++ b/packages/joint-react/rollup.config.ts
@@ -1,7 +1,7 @@
 import { createRollupConfig } from './scripts/create-rollup-config';
 
 export default createRollupConfig({
-  entries: ['src/index.ts', 'src/internal.ts', 'src/presets/index.ts'],
+  entries: ['src/index.ts', 'src/internal.ts', 'src/presets/index.ts', 'src/stories.ts'],
   external: [
     'react',
     'react-dom',
--- a/packages/joint-react/src/hooks/use-create-features.ts
+++ b/packages/joint-react/src/hooks/use-create-features.ts
@@ -277,11 +277,6 @@ export function useCreateFeature<T>(
   const graphCtx = useContext(GraphFeaturesContext);
   const featureContext = (isPaperTarget ? paperCtx : graphCtx) ?? featuresRef.current;
 
-  // Paper-specific: deferred registration when paper is not yet mounted
-  if (isPaperTarget && !featureContext.features.has(id) && !paperStore) {
-    featureContext.features.set(id, onAddFeature);
-  }
-
   const asChildren = !!paperStore;
 
   // Guard: skip onUpdateFeature on initial mount — it must only fire on dependency changes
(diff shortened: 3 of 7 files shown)


packages/joint-react/package.json
```diff
@@ -36,6 +36,11 @@
       "import": "./dist/esm/presets/index.js",
       "require": "./dist/cjs/presets/index.js"
     },
+    "./stories": {
```
samuelgja, line 39: Are we sure about this? To have stories exported?

samuelgja (approved):

PR #3307, opened 2026-05-11
fix(joint-react): controlled-mode reset preserves measured size and prunes stale cells

samuelgja:
## Description

Fixes two bugs in `@joint/react` controlled mode that surface when a parent component resets `cells` to a snapshot.

**Bug 1 — Elements collapse to `0×0` after reset.**
`mapElementToAttributes` (via `ensureDefaults`) unconditionally wrote `attributes.size = { width: 0, height: 0 }` whenever the user record omitted `size`. `applyControlled` → `graph.syncCells` → `_syncCell` then calls `cell.set(attributes)` on the existing model, merging the injected zero size and wiping the auto-measured value that the renderer had populated. The visible result is the `foreignObject` collapsing to `width=0 height=0` and only the link remaining on the paper. Same logic applied to `position`.

Fix: `ensureDefaults` now only normalises `position` / `size` when the caller actually supplied the key. Absent keys are left absent so `cell.set()` preserves whatever the existing cell already had (measured size, dragged position).

**Bug 2 — Removed cells leak back into the container after reset.**
`graph-view.ts` prune scan iterated `cells.getAll()` (the live array) while calling `cells.delete()`. The container's `delete()` swap-pops the last element into the freed slot, so the entry that lands in the freed index is skipped by the loop. Removing two stale cells in one snapshot left one of them behind, where it later surfaced as a ghost on the next non-React event.

Fix: snapshot the ids first, then delete in a separate pass.

**Pre-existing fixes pulled along the way (required for `yarn test` to pass on this branch):**
- `LinkRoutingOrthogonalOptions` was missing the `minPathMargin` field that `linkRoutingOrthogonal` destructures (`presets/link-routing.ts`).
- `rightAngleRouter(margin, minPathMargin)` made its two parameters required, but the preset/test suite called it with `0` or `1` argument. Both parameters now have defaults (`margin = 20`, `minPathMargin = margin / 4`).
- Added the missing JSDoc block description on the `usePaper` implementation overload.

## Motivation and Context

Reported during the controlled-mode docs chapter. Repro: render `<GraphProvider cells={cells} onCellsChange={setCells}>` with element records that have no `size`, add a couple of nodes, click Reset back to the initial snapshot. Old behaviour: a previously-removed task reappeared. After pulling latest: all elements collapse to `0×0` and only the link is visible.

Tests added:

- `src/state/data-mapping/__tests__/element-mapper.test.ts` — covers size-omitted, position-omitted, and partial-size normalisation.
- `src/components/graph/__tests__/graph-provider-controlled-reset.test.tsx` — covers reset preserving measured size, and stale-cell removal in a single update.

A storybook repro for the original bug report lives at `src/stories/tutorials/step-by-step/code-controlled-mode-reset-bug.tsx` (registered as `ControlledModeResetBug`).


packages/joint-react/src/presets/routers.ts
```diff
@@ -2,12 +2,17 @@ import type { routers } from '@joint/core';
...
+
 /**
  * Creates a right-angle router with `useVertices: true`.
- * @param margin
- * @param minPathMargin
+ * @param margin - distance kept from elements; defaults to `20`.
+ * @param minPathMargin - minimum path margin used by the underlying router; defaults to `margin / 4`.
  */
-export function rightAngleRouter(margin: number, minPathMargin: number): routers.Router {
+export function rightAngleRouter(
```
kumilingus: the defaults are already set in `link-routing` - this method should require both arguments.

kumilingus (approved):

PR #3315, opened 2026-05-15
feat(joint-core): originX/originY options for getFitToContentArea

kumilingus:
## Summary
`paper.getFitToContentArea` (and therefore `paper.fitToContent`) hard-anchored its grid at `(0, 0)`: grid lines landed at `0, gridWidth, 2*gridWidth, …`.

New options `originX?: number`, `originY?: number` (default `0`, paper-local coords) shift the grid anchor so lines land at `originX + n*gridWidth, originY + m*gridHeight` instead. Backward compatible — omitting the opts reproduces today's behavior exactly.

## Implementation
- `Paper.mjs::getFitToContentArea` (`src/dia/Paper.mjs`):
  - Read `opt.originX` / `opt.originY` (default `0`).
  - After scaling `area.x` / `area.y` by `sx` / `sy`, subtract `originX * sx` / `originY * sy` so the existing grid math operates in origin-relative space.
  - Return the rect translated back into absolute coords: `new Rect(-tx/sx + originX, -ty/sy + originY, …)`.
- `Paper.fitToContent` inherits the feature automatically via the existing `opt` forwarding.
- Types updated in `types/dia.d.ts` (`FitToContentOptions`).
- New unit tests cover three combinations: default origin (regression), non-zero origin with `allowNewOrigin: 'any'`, and non-zero origin without `allowNewOrigin`.

## Motivation
Consumers wanting multi-page layouts where the first page starts at a non-zero offset (e.g. a sheet preset that doesn't start at the origin) currently have no way to express it. Used by the companion `@joint/react-plus` PaperScroller refactor that introduces `sheetX` / `sheetY` props.

## Test plan
- [ ] `yarn workspace @joint/core test-client` — confirms new tests pass, no regressions.
- [ ] Backward-compat: existing `fitToContent` / `getFitToContentArea` callers behave identically when `originX` / `originY` are omitted.

### DIFF
--- a/packages/joint-core/src/dia/Paper.mjs
+++ b/packages/joint-core/src/dia/Paper.mjs
@@ -2329,6 +2329,10 @@ export const Paper = View.extend({
         const maxWidth = opt.maxWidth || Number.MAX_VALUE;
         const maxHeight = opt.maxHeight || Number.MAX_VALUE;
         const newOrigin = opt.allowNewOrigin;
+        // Shift the grid anchor so the first cell starts at (originX, originY)
+        // in paper-local coords. Default 0 reproduces the prior behavior.
+        const originX = opt.originX || 0;
+        const originY = opt.originY || 0;
 
         const area = ('contentArea' in opt) ? new Rect(opt.contentArea) : this.getContentArea(opt);
         const { sx, sy } = this.scale();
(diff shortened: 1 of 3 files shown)


kumilingus:
Superseded — origin commit cherry-picked into PR #3312 (`feat/paper-css-only-sizing`).

PR #3322, opened 2026-05-19
feat(joint-react): autoSizeOrigin on GraphProvider + identityTransform fix

kumilingus:
## Summary

- Adds `autoSizeOrigin: 'top-left' | 'center'` to `GraphProvider` (threaded to `GraphStore`). Controls which point stays fixed when a `useMeasureNode`-driven resize updates the cell.
- Removes the implicit `identityTransform` fallback in `createElementsSizeObserver`: previously echoed the cell's current x/y, which always populated `data.x/y` in the commit pipeline and short-circuited the center-anchor branch even when no user transform was supplied. Now, when `useMeasureNode` is called without a `transform`, the observer writes only `{ width, height }` and lets `autoSizeOrigin` drive any position update.
- Renames `TransformOptions.element` → `model` (and the `getCellTransform` return field) for consistency with the rest of the React API.
- Renames the internal `fromMeasure` set-option flag → `autoSize` and the resize batch name → `'auto-size'`.
- Migrates the Flowchart demo to the new option — drops the `data.cx/cy` + transform x/y math.
- New `Examples/Auto-size origin` story: two linked nodes with editable text inputs that auto-resize as you type, plus an origin toggle.

## API

```tsx
<GraphProvider autoSizeOrigin="center" initialCells={[...]}>
  <Paper />
</GraphProvider>
```

- `'top-left'` (default): element grows right/down — top-left stays put.
- `'center'`: element grows symmetrically — geometric center stays put.

Only affects measurement-driven writes. Manual `cell.resize()`, interactive resize tools, and direct `cell.set('size', ...)` calls are unaffected.

Naming mirrors CSS `transform-origin`. `'top-left' | 'center'` enum can extend to the full 9-position grid later without a breaking change.

## Flowchart demo migration

Before:
```ts
data: { label: 'Start', type: 'start', cx: 60, cy: 40 }
// + per-component transform that returned x = cx - W/2, y = cy - H/2
```

After:
```ts
data: { label: 'Start', type: 'start' }, position: { x: 60, y: 40 }
// transform only adds padding; autoSizeOrigin="center" anchors the resize
```

JointJS default element size is 0×0, so `position` doubles as the desired center on first render — `autoSizeOrigin="center"` then preserves that center through the first DOM measurement.

## No-transform path

`createElementsSizeObserver` previously fell back to an `identityTransform` when `useMeasureNode` had no user `transform`. The fallback re-emitted the cell's current x/y, which then hit the explicit-position branch in `GraphStore.onBatchUpdate` and never reached the new center-anchor branch. The fallback is now gone: with no user transform, the observer writes `{ width, height }` only, and `autoSizeOrigin` decides what (if anything) happens to position.

## Renames

- `TransformOptions.element` → `TransformOptions.model` (also in `getCellTransform`'s return shape).
- Internal `fromMeasure` set-option → `autoSize`; batch name `'resize'` → `'auto-size'`. These are private to the React store and not part of the public API.

## Test plan

- [ ] `yarn workspace @joint/react typecheck`
- [ ] `yarn workspace @joint/react lint`
- [ ] `yarn workspace @joint/react jest --testPathPatterns="graph-store|graph-provider|size-observer|measure-node"` — all pass
- [ ] Storybook: open `Examples/Auto-size origin` — type into the inputs, toggle origin, verify both modes anchor as documented
- [ ] Storybook: open `Demos/Flowchart` — verify layout matches the previous version (no shift)

### DIFF
--- a/packages/joint-react/src/store/graph-store.ts
+++ b/packages/joint-react/src/store/graph-store.ts
@@ -227,16 +227,13 @@ export class GraphStore<
         this.graph.stopBatch('auto-size');
       },
       getCellTransform: (id) => {
-        const cell = this.graph.getCell(id);
-        if (!cell?.isElement()) throw new Error('Cell not valid');
-        const size = cell.size();
-        const position = cell.get('position');
+        const model = this.graph.getCell(id);
+        if (!model?.isElement()) throw new Error('Cell not found or not an element: ' + id);
         return {
-          width: size.width,
-          height: size.height,
-          element: cell,
-          angle: cell.get('angle') ?? 0,
-          ...position,
+          model,
+          ...model.size(),
+          ...model.position().toJSON(),
+          angle: model.angle(),
         };
       },
     });
--- a/packages/joint-react/src/store/graph-store.ts
+++ b/packages/joint-react/src/store/graph-store.ts
@@ -51,6 +51,18 @@ export interface GraphStoreInternalSnapshot {
  * to rule out the nonsensical combination of `cells` and `initialCells`
  * being passed together.
  */
+/**
+ * Reference point that stays fixed when an auto-sized element's measured size
+ * changes. Mirrors CSS `transform-origin` semantics.
+ *
+ * - `'top-left'` (default): element grows right/down — top-left stays put.
+ * - `'center'`: element grows symmetrically — geometric center stays put.
+ *
+ * Only affects writes from the `useMeasureNode` pipeline. Manual `cell.resize()`,
+ * interactive resize tools, and direct `cell.set('size', ...)` calls are unaffected.
+ */
+export type AutoSizeOrigin = 'top-left' | 'center';
+
 interface GraphStoreOptionsBase<
   Element extends ElementJSONInit = ElementJSONInit,
   Link extends LinkJSONInit = LinkJSONInit,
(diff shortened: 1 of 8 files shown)


packages/joint-react/src/store/graph-store.ts
```diff
@@ -179,20 +201,28 @@ export class GraphStore<
         this.graph.startBatch('resize');
```
kumilingus: can rename it to `auto-size`?

packages/joint-react/src/store/graph-store.ts
```diff
@@ -179,20 +201,28 @@ export class GraphStore<
...
-            if (data.x !== undefined && data.y !== undefined) {
-              cell.set('position', { x: data.x, y: data.y }, {
-                fromMeasure: true,
-              } as object);
-            }
+          if (!cell?.isElement()) continue;
+          // Capture center BEFORE size write so the center-anchor path can
+          // re-derive position from the pre-resize center.
+          const center = this.autoSizeOrigin === 'center' ? cell.getCenter() : null;
+          const setOptions = { fromMeasure: true };
```
kumilingus: can we change it to `autoSize: true` ?

PR #3325, opened 2026-05-20
feat(joint-react): rename PortalPaper → ReactPaper, fix is-measuring stuck class, story cleanup

kumilingus:
### Rename \`PortalPaper\` → \`ReactPaper\`
The custom \`dia.Paper\` subclass that wires JointJS into React's render lifecycle is renamed to better reflect its purpose. Public-API impact:

- \`export { ReactPaper }\` from \`@joint/react\` (was \`PortalPaper\`).
- \`usePaper()\` now returns \`{ paper: ReactPaper }\` instead of \`{ paper: dia.Paper }\` — consumers get the subclass type with React-specific methods (\`FLAG_MEASURE\`, etc.) accessible without casts.
- Model-side options type renamed: \`PortalPaperOptions\` → \`ReactPaperOptions\` (file \`models/react-paper.types.ts\`).
- Source/test files renamed: \`models/portal-paper.{ts,types.ts}\` → \`models/react-paper.{ts,types.ts}\`; matching test files.

**Kept intentionally** (different concept, not the Paper class itself):
- \`useCreatePortalPaper\` hook + \`use-create-portal-paper.tsx\` (portal-rendering hook).
- \`PortalHostCell\`, \`PortalSelector\`, \`PortalSelectorContext\` types.
- The \`PortalPaperOptions\` interface in \`components/paper/paper.types.ts\` (separate interface — public Paper component props subset, not the class options).

### Bug fix: \`.jj-is-measuring\` stuck when measurement matches the model
Previously \`PortalPaper.updateView\` stripped \`MEASURING_CLASS_NAME\` on the next updateView for an element view. That works when measurement causes a size/position change, but \`processSizeChange\` short-circuits when the measured size already matches the model (within \`EPSILON\`) — no \`onBatchUpdate\`, no \`updateView\`, the class stayed forever.

Fix: route the reveal through an explicit \`FLAG_MEASURE\` update.
- \`ReactPaper\` declares \`FLAG_MEASURE = 1 << 27\` and only clears the class when this flag is present on the incoming update.
- \`useMeasureNode\` calls \`paper.requestViewUpdate(view, FLAG_MEASURE, …)\` right after adding the class, so the next update cycle reliably removes it regardless of whether the model size changed.

Added regression test in \`use-measure-node.test.tsx\` with a local \`TestResizeObserver\` mock that delivers entries (the package-level setup installs a stub): cell sized 80×120, rect rendered at 80×120, observer fires with the same dims, asserts the class is gone.

### Story renames + cleanup
- Folder renames: \`with-build-in-shapes\` → \`with-native-shapes\`; \`with-element-ports-native\` → \`with-native-ports\`.
- Story title renames: "Built-in Shapes" → "Native shapes"; "Element Ports (Native)" → "Native ports".
- Refresh the \`with-native-shapes\` story description to match the current \`CellRecord\` + \`initialCells\` API (drops the stale \`mapDataToElementAttributes\` reference).
- Delete the obsolete \`with-custom-mapper\` example.

## Test plan
- [x] \`yarn typecheck\`
- [x] \`yarn jest\` — 865 pass; only the pre-existing \`graph-store-features\` \`/Cell not valid/\` failure remains (lives on \`dev\`, unrelated)
- [ ] \`yarn storybook\` — confirm renamed stories load under their new titles
- [ ] Visual: native-shapes story renders every standard shape and link variant
- [ ] Visual: native-ports story renders ports with magnets
- [ ] Visual: element with \`useMeasureNode\` whose initial size already matches its rendered DOM size becomes visible after the first paper update cycle


packages/joint-react/src/components/paper/paper.types.ts
```diff
@@ -1,7 +1,7 @@
 import type { dia } from '@joint/core';
 import type { LinkRecord } from '../../types/cell.types';
-import type { PortalSelector } from '../../models/portal-paper.types';
-import type { PortalPaper } from '../../models/portal-paper';
+import type { PortalSelector } from '../../models/react-paper.types';
+import type { ReactPaper } from '../../models/react-paper';
```
samuelgja, line 4: hehe

samuelgja (approved):

PR #3326, opened 2026-05-21
fix(joint-react): useCells no-arg form not reactive to cell data changes

samuelgja:
## Description

The no-arg `useCells()` form never re-rendered when cell data changed because
`container.getAll()` returns the same mutable array reference. Both `Object.is`
and `areArraysShallowEqual` short-circuited on `a === b`, so React never detected
changes.

Two fixes:
1. `isEqualCallback` now uses `areArraysShallowEqual` for all array-returning
   forms without selectors (previously missed the no-args case).
2. The `select` function shallow-copies the raw container array so the cached
   value is a distinct reference, enabling element-by-element comparison on
   subsequent calls.

Copy only happens in `useCells` — `container.getAll()` still returns the raw
mutable array for iteration-only callers (no overhead).

## Motivation and Context

Inspector panels (e.g. element-ports story) using `useCells()` to read all cells
did not update when cell properties changed via `setCell`. This affected any
component subscribing to the full cells array without a selector.

### Regression tests

- `no-arg form re-renders when a cell data changes` — verifies the fix
- `no-arg form handles large cell counts without stack overflow` — 5000-cell stress test


packages/joint-react/src/store/state-container.ts
```diff
@@ -78,9 +72,7 @@ export function asReadonlyContainer<
  * not rely on insertion order for identity.
  * @param _name - optional label for debugging
  */
-export function createContainer<
-  Cell extends AnyCellRecord,
->(_name?: string): Container<Cell> {
+export function createContainer<Cell extends AnyCellRecord>(_name?: string): Container<Cell> {
```
kumilingus: `_name` is unused

PR #3334, opened 2026-05-27
fix(joint-react): discard stale feature ref after StrictMode cleanup in useCreateFeature

Geliogabalus:
## Summary

- Fixes `useCreateFeature` re-registering a dead feature reference after React StrictMode double-invoke unmounts it
- In StrictMode the cleanup runs `feature.clean()`, which destroys the underlying JointJS views. The cached `featureRef.current` still pointed to that cleaned-up instance, causing re-registration of broken views on the second mount
- Now checks whether the cached ref is still present in the store via `resolveExistingFeature`; if it was evicted (StrictMode cleanup path), the ref is discarded so the normal creation path below builds a fresh instance instead

## Test plan
- [ ] Verify features (e.g. highlighters, tools) mount and render correctly in StrictMode (`<React.StrictMode>`)
- [ ] Confirm no "dead view" errors or visual glitches on second StrictMode mount
- [ ] Confirm features still work correctly outside StrictMode (single mount, no regression)

### DIFF
--- a/packages/joint-react/src/hooks/use-create-features.ts
+++ b/packages/joint-react/src/hooks/use-create-features.ts
@@ -331,12 +331,21 @@ export function useCreateFeature<T>(
   useLayoutEffect(() => {
     if (isPaperTarget && !paperStore) return;
 
-    // Re-register cached feature if it was removed by strict-mode cleanup
-    // OR if it was already created during render (inside-Paper case).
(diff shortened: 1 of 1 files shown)


Geliogabalus:
In StrictMode Selection and Snaplines components were not working in a standalone application (after building). Claude suggested this change which works.

samuelgja (approved):

PR #3336, opened 2026-05-29
refactor(joint-react): default paper IDs + unified usePaper/usePaperStore signatures

samuelgja:
## Description

Streamlines `<Paper>` identity and the paper-access hooks:

- **Default paper id** — `<Paper>` without an `id` registers under `DEFAULT_PAPER_ID`; every paper-id hook argument is now optional and resolves **context → default paper**. Duplicate default-paper registrations emit a dev warning (`dev-warnings.ts`).
- **Unified signatures** — `usePaperStore(id?)` and `usePaper(id?)` collapse their overloads into one signature; view access is typed `T | undefined` (never throws), and the returned object is always stable. Removed the `OPTIONAL` / `Optional` sentinel.
- **Result types → `[Concept]Handle`** — `UseGraphResult → GraphHandle`, `UsePaperResult → PaperHandle`, `UseCreatePortalPaperResult → CreatePortalPaperHandle`.
- **Dedup / utils** — extracted `resolvePaper` / `resolvePaperId` / `isPaperTarget` into `utils/resolve-paper-target.ts` (+ tests) and shared `makeOptions` / `assignOptions` / `pickValues` in `utils/object-utilities.ts`; trimmed `internal.ts` to what `@joint/react-plus` actually consumes.
- Cleaned hardcoded paper IDs and unused hooks across demo / example stories.

## Motivation and Context

The paper hooks had overloads that flipped return types between guaranteed and optional, and some threw while others returned `null`. This unifies the contract — views are `undefined` until mounted, models throw on misconfiguration — and lets a single `<Paper>` work without an explicit `id`, improving DX.

### DIFF
--- a/packages/joint-react/src/hooks/use-measure-node.tsx
+++ b/packages/joint-react/src/hooks/use-measure-node.tsx
@@ -176,8 +176,8 @@ export function useMeasureNode(
     // Hide the element view's until the element view is updated by the paper's
     // update cycle. This prevents flashes of incorrect position when the element
     // is measured, but waits for the paper to apply the new position.
-    const view = paper.findViewByModel(id);
-    if (view) {
+    const view = paper?.findViewByModel(id);
+    if (view && paper) {
       view.el.classList.add(MEASURING_CLASS_NAME);
       // Request a view update to remove the measuring class as part of its
       // next update cycle.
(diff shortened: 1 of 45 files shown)


packages/joint-react/src/hooks/use-measure-node.tsx
```diff
@@ -176,8 +176,8 @@ export function useMeasureNode(
     // Hide the element view's until the element view is updated by the paper's
     // update cycle. This prevents flashes of incorrect position when the element
     // is measured, but waits for the paper to apply the new position.
-    const view = paper.findViewByModel(id);
-    if (view) {
+    const view = paper?.findViewByModel(id);
+    if (view && paper) {
```
kumilingus, line 180: unnecessary.

PR #3359, opened 2026-06-12
refactor(joint-react): rename measurement hooks + DOM ref convention

kumilingus:
## Summary

- **Rename `useMeasureNode` → `useMeasureElement`** (and `MeasureNodeOptions` → `MeasureElementOptions`). Hook already throws when called on a link — name should match. Aligns with the `Element` terminology used everywhere else (`ElementModel`, `ElementRecord`, `selectElementSize`).
- **Rename `useNodesMeasuredEffect` → `useOnElementsMeasured`.** It's an event subscription, not an effect; `useOn*` reads better. Also **drops the `dependencies` argument** from both overloads — the hook already uses a `callbackRef.current = callback` pattern so re-subscription on callback identity change was wasteful and not part of any documented contract.
- **Rename DOM refs `elementRef` → `nodeRef`** in code samples, stories, and internal infra. `element` is the diagram item; `node` is the DOM node — using `elementRef` for a DOM ref was confusing.
- **Rename Storybook example** `Examples/Element Defaults` (`stories/examples/element-defaults/`) → `Examples/Custom Cell Models` (`stories/examples/custom-cell-models/`).

Two existing story call sites that passed a `dependencies` array to `useNodesMeasuredEffect` are updated to drop it: `stories/examples/automatic-layout/code.tsx`, `stories/demos/automatic-layout-storage/code.tsx`.

## Test plan

- [ ] `yarn typecheck` clean
- [ ] `yarn docs:typedoc` reports 0 warnings
- [ ] `yarn jest src/hooks` — all suites pass
- [ ] Storybook builds and the renamed `Examples/Custom Cell Models` story renders


packages/joint-react/.storybook/decorators/with-simple-data.tsx
```diff
@@ -133,12 +133,12 @@ export function SimpleRenderLinkDecorator(Story: StoryFunction, { args }: StoryC
 }
 
 export function HTMLNode(props: PropsWithChildren<HTMLProps<HTMLDivElement>>) {
-  const elementRef = useRef<HTMLDivElement>(null);
-  const { width, height } = useMeasureNode(elementRef);
+  const nodeRef = useRef<HTMLDivElement>(null);
```
samuelgja: this naming is better "divRef"

samuelgja (approved):

PR #3360, opened 2026-06-12
chore(joint-react): update package.json [dev]

zbynekstara:
## Description

Updates package.json of joint-react to match other packages.


samuelgja (approved):

samuelgja (approved):
I think we need make a fix that we need move to the dist css files. But i think in separate PR!

samuelgja (approved):

kumilingus (approved):

PR #3366, opened 2026-06-15
fix(joint-react): interactive={true} disables linkMove/labelMove by default

samuelgja:
## Description

Changes how the `interactive` Paper prop maps to native JointJS:

- `interactive={false}` → `false` (every interaction disabled)
- `interactive={true}` → `{ linkMove: false, labelMove: false }` (rest implicit `true`)
- object form → defaults merged, user keys win
- function form → explicit takeover (unchanged)

Connected elements stay movable under `interactive={true}` (link endpoints follow); only link-body drag (`linkMove`) and label drag (`labelMove`) are off by default.

## Motivation and Context

`interactive={true}` previously passed raw `true` to JointJS, enabling `linkMove`/`labelMove` — links and labels dragged unintentionally. Aligns boolean handling with the documented default (`labelMove: false`) plus `linkMove: false`.


samuelgja:
wrong base

PR #3383, opened 2026-06-19
feat(joint-react): generic cellVisibility ownership latch

kumilingus:
## Why

In `@joint/react-plus`, using `<Paper cellVisibility>` together with `<PaperScroller virtualRendering>` throws:

> `VirtualRenderingController: Paper "cellVisibility" is already set. The controller cannot override it.`

The `VirtualRenderingController` needs to own `paper.options.cellVisibility` (it composes the user callback with a viewport check), but the React `<Paper>` writes that option on creation **and** re-asserts it on every relevant re-render (a passive effect that always runs after the scroller's layout effect). So the scroller's wrapper gets clobbered.

This can't be fixed from the plus layer alone — core must stop writing `cellVisibility` when a feature owns it.

## What

A **generic** ownership latch on `PaperStore`. Core has **no** knowledge of PaperScroller or virtual rendering — any feature can claim the option through the existing feature system.

- `PaperStore`:
  - `nativeCellVisibility` — the resolved native callback (kept current regardless of ownership).
  - `claimCellVisibility(ownerId)` / `releaseCellVisibility(ownerId)` / `isCellVisibilityOwned`.
  - `registerCellVisibilityListener` / `notifyCellVisibilityChange` — push refreshed callbacks to the owner without depending on the owning component re-rendering.
- `use-create-portal-paper`: expose `nativeCellVisibility`; when owned, skip writing `cellVisibility` onto the paper and notify the owner instead.

## Tests

`paper-store.test.ts` — 7 new cases (claim clears, release restores, non-owner no-op, listener fired/cleared). Full suite: **946 passing**.

## Coordinated change

Consumed by a matching PR in `clientIO/joint-plus` (PaperScroller + VirtualRenderingController). The plus side feature-detects this latch and degrades gracefully against an older core.

### DIFF
--- a/packages/joint-react/src/components/paper/__tests__/paper.test.tsx
+++ b/packages/joint-react/src/components/paper/__tests__/paper.test.tsx
@@ -67,4 +67,22 @@ describe('Paper', () => {
       expect(refHolder.current).not.toBeNull();
     });
   });
+
+  it('throws when cellVisibility is set via the options escape hatch', () => {
+    // `PaperOptions` excludes `cellVisibility` at the type level; cast around
+    // it to simulate a plain-JS caller and assert the runtime guard.
+    const spy = jest.spyOn(console, 'error').mockImplementation(() => {});
+    expect(() =>
+      render(
+        <GraphProvider initialCells={CELLS}>
+          <Paper
+            style={{ width: 100, height: 100 }}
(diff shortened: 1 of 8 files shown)


kumilingus:
Consumer PR: clientIO/joint-plus#715

PR #3401, opened 2026-07-01
feat(react): add useGraphTransaction — atomic graph edits with single-render coalescing + rollback

samuelgja:
## Description

Adds `useGraphTransaction()` (also exposed as `useGraph().transaction`) — run many graph edits as one atomic unit:

- **One undo entry + one React update** for a batch of edits. Sync edits already coalesce via the scheduler; this also makes **async** transactions (edits split across `await`s) coalesce to a **single** React update, by deferring the container commit while a `'transaction'` batch is open and flushing once on close. The deferral is gated to the `'transaction'` batch name, so drags / layout / auto-size / controlled `updateFromReact` are unchanged.
- **Automatic rollback** on throw/reject (opt out with `{ rollback: false }`). The snapshot is a shallow `[...cells.getAll()]` — container records are immutable (slots are replaced, never mutated), so a shallow copy is a correct and cheap pre-transaction snapshot; restore is `resetCells(snapshot)`.
- Works in **uncontrolled and controlled** mode (verified: an async 3-edit transaction fires `onCellsChange` / commits exactly once).
- Callback may be sync or `async`; async is awaited before the transaction closes and the call returns the pending promise.

Also:
- `isPromise` added to `utils/is.ts`.
- `DEFAULT_PAPER_ID` is now exported from the public API.

## Motivation and Context

Grouping edits previously meant reaching for the plus-only `CommandManager.initBatchCommand()/storeBatchCommand()`. This gives `@joint/react` a first-class, framework-level primitive that also guarantees a single re-render (even across `await`s) and provides atomic rollback — without depending on `@joint/plus`.

N/A

### DIFF
--- a/packages/joint-react/stories/examples/transaction/docs.mdx
+++ b/packages/joint-react/stories/examples/transaction/docs.mdx
@@ -1,0 +1,66 @@
+import { Meta, Canvas, Markdown } from '@storybook/addon-docs/blocks';
+import * as Stories from './story';
+import Code from './code?raw';
+
+<Meta of={Stories} />
+
+# Transactions
+
+`useGraph().transaction` groups any number of graph edits into one atomic step.
+
+```tsx
+const { transaction, setCell } = useGraph();
+
+transaction(() => {
+  for (const id of ids) setCell(id, (prev) => ({ ...prev, position: layout(id) }));
+});
+```
+
+Three things happen for the whole block, not per edit:
+
+- **One undo entry.** Every write collapses into a single command (grouped via a JointJS batch).
+- **One React update.** Synchronous edits already coalesce; a transaction also coalesces **async** edits split across `await`s into a single commit.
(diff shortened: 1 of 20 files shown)


packages/joint-react/stories/examples/transaction/docs.mdx
```diff
@@ -0,0 +1,66 @@
...
+${Code}
+\`\`\``}
+</Markdown>
+
+## Async transactions
+
+The callback may be `async`. It is awaited before the transaction closes, and the call returns the pending promise, so you can `await` the whole unit:
+
+```tsx
+await tx(async () => {
```
kumilingus, line 45: tx -> transaction

packages/joint-react/stories/examples/transaction/docs.mdx
```diff
@@ -0,0 +1,66 @@
...
+});
+```
+
+## Rollback
+
+Rollback snapshots the cells before running and restores them on failure, so a rejected transaction leaves the graph exactly as it was:
+
+```tsx
+try {
+  await tx(async () => {
```
kumilingus, line 59: tx -> transaction

kumilingus (approved):

PR #3412, opened 2026-07-05
types(joint-react): useCells/useCell generic is the record, returns Computed (#3412)

kumilingus:
## Summary

`useCells` / `useCell` had their `Cell` type parameter meaning the **resolved** return shape, defaulting to `Computed<CellRecord>`. To get typed reads you had to write the awkward `useCells<Computed<MyCell>>()`.

This flips the generic so `Cell` is the **input record** (default `CellRecord`) and the hooks return its `Computed` form:

```ts
// before
const cells = useCells<Computed<WorkflowCell>>(collection);
// after
const cells = useCells<WorkflowCell>(collection); // → Computed<WorkflowCell>[]
```

This matches how `useGraph<Element, Link>` already takes record types, and removes the `Computed<>` boilerplate at call sites.

## Backward compatibility

`Computed` is **idempotent** (`Computed<Computed<T>> = Computed<T>`) and distributes over unions, so existing `useCells<Computed<MyCell>>()` call sites resolve to the same type — no breakage. Marked `!` because the generic's *meaning* changes (a custom resolved type not shaped like `ElementRecord`/`LinkRecord` would now be re-mapped by `Computed`).

## Scope

- Types only — no runtime change.
- `packages/joint-react/src/hooks/use-cells.ts` (8 overloads + impl), `use-cell.ts` (4 overloads + impl). Internal helpers still operate on the resolved type; only the public surface wraps in `Computed<Cell>`.
- All hook tests pass **unchanged**, including the type tests (`use-cells.type.test.ts`, `use-cell.type.test.ts`) — the idempotency keeps them valid.

## Verification

- `yarn typecheck` ✅
- `yarn lint` ✅
- `yarn jest` ✅ (1023 tests)

### DIFF
--- a/packages/joint-react/src/hooks/__tests__/use-cell.type.test.ts
+++ b/packages/joint-react/src/hooks/__tests__/use-cell.type.test.ts
@@ -42,6 +42,8 @@ interface LinkUserData {
 }
 type MyElement = Computed<ElementRecord<ElementUserData>>;
 type MyLink = Computed<LinkRecord<LinkUserData>>;
+// Bare (write-shape) record; the hook must resolve it to its `Computed` form.
+type MyElementRecord = ElementRecord<ElementUserData>;
 
 /** Compile-time assertion that `actual` is assignable to `Expected`. */
 const expectType = <Expected>(_actual: Expected): void => {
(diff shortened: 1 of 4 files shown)


samuelgja (changes requested):
Does a checks pass? Yarn test (typecheck, eslint, jest)?  Becasue there were dedicated "type" tests for this useCell, so if that pass, then its okey.

samuelgja (approved):

PR #3436, opened 2026-07-24
chore: enable yarn deferred versioning

zbynekstara:
## Description

Adopt [Yarn's deferred versioning](https://yarnpkg.com/features/release-workflow) so the intended release type (patch / minor / major) for each publishable `@joint/*` package is declared **per pull request** instead of being hand-applied at release time.

Contributors run `yarn version-check` and commit a small `.yarn/versions/*.yml` file describing their bump; CI blocks PRs that change a publishable package without one. At release a maintainer runs `yarn version-apply` to apply all accumulated bumps at once (Yarn rewrites cross-package `workspace:` ranges automatically).

## Motivation and Context

Releases are currently fully manual: a maintainer hand-edits each `package.json` `version` and the `CHANGELOG`, then tags and publishes. Nothing records what kind of release a normal PR implies, so the bump decision is deferred to one person at the end and is easy to get wrong. Deferred versioning captures that intent at PR time, i.e. by the author with the most context, and makes the release step mechanical.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -3,10 +3,11 @@ Thank you for submitting a pull request!
 
 Please verify that:
 * [ ] Code is up-to-date with the `master` branch
-* [ ] You've successfully run `grunt test` locally
+* [ ] You've successfully run `yarn test` locally
 * [ ] If applicable, there are new or updated unit tests validating the change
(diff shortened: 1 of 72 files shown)


zbynekstara:
Replaced by changesets

PR #3442, opened 2026-08-02
test(joint-react): add repro for useCell throw on cell removal

kumilingus:
Adds a minimal Storybook reproduction under `Bugs/Controlled cells: removing a cell`. No library code is touched.

## The bug

Removing a cell from a controlled `cells` array throws and takes the whole paper down:

```
Error: useCell(): no cell with id "b"
    at use-cell.ts:31
    at computeNext (use-cells.ts:51)
```

The trigger is a `renderElement` subtree that **subscribes** to its own cell. When the cell leaves the store, React re-runs that subtree's selector before the parent reconciles the removal, so the selector looks up a cell that is already gone.

Renaming a node is the everyday way in: an id is often authored data — a node name in a text source — so editing that name is a remove plus an add. The repro came out of a Mermaid-to-JointJS renderer where every keystroke in a node id crashed the canvas.

`useCell` is called directly in the story to keep it small, but `useMeasureElement` reaches the same line via `useCell(selectElementSize)`, which is how a content-sized node hits this without ever naming `useCell`.

## Stories

Press the button in each; measured with Playwright against `storybook dev`:

| Story | Elements before → after | Result |
| --- | --- | --- |
| `Broken` | 2 → 0 | throws `useCell(): no cell with id "b"` |
| `WithoutSubscription` | 2 → 2 | no throw |
| `WorkaroundRemountOnIdChange` | 2 → 2 | no throw |

`WithoutSubscription` is the narrowing result: identical cell updates, but `renderElement` reads only its `data` argument. It survives — so the subscription is the necessary ingredient, not the removal on its own.

`WorkaroundRemountOnIdChange` keys the provider on the set of ids, which unmounts the subscribed subtrees before the new cells are applied. It works, at the cost of a full graph rebuild on every id change.

## Notes

- Reproduces on `4.3.1` (this tree) and the published `4.3.2` — `use-cell.ts` has the same unguarded throw in both.
- A wholesale replacement of every id fails too, so this is not specific to a partial diff.


samuelgja:
https://github.com/clientIO/joint/pull/3447 close by favor of new pr!

PR #3478, opened 2026-08-23
fix(joint-react): do not hide links connected to other links

kumilingus:
## Problem

The react paper hides a freshly inserted link view (`visibility: hidden`) until its end views are "ready" — portal-rendered elements are empty until their React content mounts, so a link pointing at them would otherwise dangle in the void.

The readiness check (`isLinkEndReady → isElementReady → getElementView(id)`) only ever resolves the end id among **elements**. When a link's end is another **link** (e.g. a BPMN annotation attached to a flow), the end never becomes ready and the link stays hidden forever — `checkPendingLinks` re-checks with the same test on every element insert, so it never recovers.

## Fix

A link end never needs to be waited for — only elements render asynchronously. `isLinkEndReady` now returns `true` when the end cell is a link.

Elements without portals are unaffected (already handled per view by `isElementReady` — a view with no portal node is ready), so mixed portal/portal-less diagrams keep working.

## Tests

- link connected to another link is not hidden
- a chain of link-to-link connections is not hidden
- links between portal-less elements are not hidden

### DIFF
--- a/packages/joint-react/src/mvc/paper.ts
+++ b/packages/joint-react/src/mvc/paper.ts
@@ -128,11 +128,18 @@ export class PaperView extends Paper {
   /**
    * Check whether a link end can be rendered immediately.
    * @param end - The link end JSON descriptor.
-   * @returns True if the link end's target element is ready.
+   * @returns True if the link end's target cell is ready.
    */
   private isLinkEndReady(end: dia.Link.EndJSON): boolean {
-    if (!end.id) return true;
-    return this.isElementReady(end.id as CellId);
+    const endId = end.id as CellId | undefined;
+    if (!endId) return true;
+
+    // Only elements can render asynchronously (their React portal content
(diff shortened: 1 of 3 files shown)


samuelgja (approved):

packages/joint-react/src/mvc/paper.ts
```diff
@@ -128,11 +128,18 @@ export class PaperView extends Paper {
   /**
    * Check whether a link end can be rendered immediately.
    * @param end - The link end JSON descriptor.
-   * @returns True if the link end's target element is ready.
+   * @returns True if the link end's target cell is ready.
    */
   private isLinkEndReady(end: dia.Link.EndJSON): boolean {
-    if (!end.id) return true;
-    return this.isElementReady(end.id as CellId);
+    const endId = end.id as CellId | undefined;
```
samuelgja: I believe this type assertion is not necessary, or is it?

PR #3484, opened 2026-08-25
fix(joint-react): keep the diagram rendered and wired across dev-server hot reloads (HMR)

### LINKED ISSUE #3483: react dev server (hot reload) issue

samuelgja:
## Description

Hot reloading on a dev server (Vite / any Fast Refresh HMR) could leave the diagram blank or unresponsive until a full page reload. Three root causes, all fixed:

1. **`GraphProvider` kept publishing a destroyed store.** A Fast Refresh re-runs all effects of the refreshed component (dependency arrays are ignored), so the provider's mount effect destroys and re-creates its `GraphStore` — but readiness was tracked as a boolean, so `setIsReady(true)` bailed out and consumers never learned about the new instance: the context kept handing out the destroyed store (cleared graph, dead subscriptions). `useImperativeApi()` now keeps the instance itself in React state and exposes it; the context value and the provider wiring effect are keyed on the instance identity, so a re-created store is published and re-wired (controlled `cells` re-applied included).

2. **`<Paper>` never re-registered against a swapped store.** The paper registration effect had `[]` dependencies, so it stayed bound to the destroyed store forever. It is now keyed on `[graphStore, id]`, so a store re-created while the paper stays mounted gets a fresh registration (and a paper `id` change re-registers as well).

3. **`instanceof PaperView` guards blanked all portals after a module re-evaluation.** HMR re-evaluates `mvc/paper.ts`, producing a new `PaperView` class identity while live paper instances still come from the previous evaluation; the element/link portal guards rejected them and every portal rendered `null`. The guards now use a capability check (`isPaperView()`), which survives re-evaluation. This is the StrictMode-flavored variant of the bug: under StrictMode the strict double-render replay resets `ignorePreviousDependencies`, so the effect re-run path never fires and only the class-identity hazard remains.

**Known behavior, unchanged by this fix (uncontrolled mode):** a hot reload that re-creates the provider's store re-seeds the graph from `initialCells`, so runtime-only edits (cells added at runtime, dragged positions) reset — the same as a remount. This is inherent to the provider owning the store inside an effect. Controlled mode (`cells`) is re-applied to the new store, and an external `graph` prop is adopted without being cleared, so both preserve state across the reload. This PR targets the blank/broken canvas, not state round-tripping.

The regression tests drive the **real `react-refresh` runtime against the real `react-dom`**: the suite injects the refresh hook before `react-dom` loads, renders, re-evaluates only `graph-provider.tsx` with all its dependencies shared (the same module-graph shape a Vite HMR update produces), runs `performReactRefresh()`, and asserts the diagram stays rendered, does not remount, and remains live (an imperative `graph.addCell()` still paints). A second suite pins the cross-evaluation `PaperView` guard. `react-refresh` is added as a dev dependency for these tests.

Fixes #3483

## Motivation and Context

Editing library source with the storybook/Vite dev server running would "sometimes" — depending on which module the edit invalidated — blank the canvas or freeze subscriptions, forcing a full page reload. This makes HMR reliable in both non-StrictMode apps (effect re-run path) and StrictMode apps (class-identity path).

### DIFF
--- a/packages/joint-react/src/mvc/paper.ts
+++ b/packages/joint-react/src/mvc/paper.ts
@@ -20,6 +20,23 @@ export const DEFAULT_PAPER_ID = 'default-paper';
  * - Emitting view mount/unmount callbacks for graph-store snapshot sync
  * - Hiding links until their source/target elements have rendered
  */
+/**
+ * Type guard: is this paper a {@link PaperView}?
+ *
+ * Checks the {@link PaperView.getCellViewPortalNode} capability instead of
+ * `instanceof`: a dev-server hot reload (HMR) can re-evaluate this module and
+ * create a new `PaperView` class identity while live paper instances still
+ * come from the previous evaluation — `instanceof` would reject those and
+ * blank the canvas.
+ * @param paper - The paper instance to check (nullish tolerated).
+ * @returns `true` when the paper exposes the `PaperView` portal contract.
+ */
+export function isPaperView(paper: dia.Paper | null | undefined): paper is PaperView {
+  return (
+    !!paper && 'getCellViewPortalNode' in paper && typeof paper.getCellViewPortalNode === 'function'
+  );
+}
+
 export class PaperView extends Paper {
   public viewChanges: Map<CellId, IncrementalChange<dia.Cell>> = new Map();
   public onViewMountChange: (changes: Map<CellId, IncrementalChange<dia.Cell>>) => void;
--- a/.changeset/react-hmr-graph-store.md
+++ b/.changeset/react-hmr-graph-store.md
@@ -1,0 +1,5 @@
+---
+"@joint/react": patch
+---
+
+<GraphProvider /> - fix the diagram going blank when a dev-server hot reload (HMR) re-creates the graph store
--- a/.changeset/react-hmr-paper-view-guard.md
+++ b/.changeset/react-hmr-paper-view-guard.md
@@ -1,0 +1,5 @@
+---
+"@joint/react": patch
+---
+
+<Paper /> - fix element and link portals vanishing when a hot reload re-evaluates the `PaperView` class
--- a/packages/joint-react/package.json
+++ b/packages/joint-react/package.json
@@ -123,6 +123,7 @@
     "react": "^19.1.1",
     "react-dom": "^19.1.1",
     "react-redux": "^9.2.0",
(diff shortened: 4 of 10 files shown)


packages/joint-react/src/mvc/paper.ts
```diff
@@ -13,6 +13,23 @@ const noopViewMountChange = (): void => {
...
+ * `instanceof`: a dev-server hot reload (HMR) can re-evaluate this module and
+ * create a new `PaperView` class identity while live paper instances still
+ * come from the previous evaluation — `instanceof` would reject those and
+ * blank the canvas.
+ * @param paper - The paper instance to check (nullish tolerated).
+ * @returns `true` when the paper exposes the `PaperView` portal contract.
+ */
+export function isPaperView(paper: dia.Paper | null | undefined): paper is PaperView {
+  return (
+    !!paper && 'getCellViewPortalNode' in paper && typeof paper.getCellViewPortalNode === 'function'
```
kumilingus: Could we add a marker symbol on the paper view class and check based on that?

See e.g. https://github.com/clientIO/joint/blob/master/packages/joint-core/src/dia/GraphLayer.mjs#L13

kumilingus (changes requested):

.dev-worktree
kumilingus, line 1: This should not be here.

kumilingus (changes requested):

packages/joint-react/src/mvc/paper.ts
samuelgja: Done: `PaperView` now carries a marker symbol as a class field (`readonly [PAPER_VIEW_MARKER] = true`, same shape as `GRAPH_LAYER_MARKER`) and `isPaperView()` checks that. One deliberate difference: the symbol comes from `Symbol.for('joint.react.paperViewMarker')` rather than `Symbol()`, because an HMR update re-evaluates this module and a module-local symbol would differ between the class that created the live papers and the one doing the check, the same hazard `instanceof` has. The test now builds real instances from both evaluations of the class.

.dev-worktree
samuelgja: Removed. It was the gitlink of a local git worktree directory that got added by accident; `git rm --cached .dev-worktree` is staged and the path is excluded locally so it cannot come back.

kumilingus (approved):

PR #3497, opened 2026-09-07
feat(joint-layout-elk): new package - ELK layout with container and port support

Geliogabalus:
## Summary

Introduces `@joint/layout-elk`, a new package that lays out JointJS graphs using the [Eclipse Layout Kernel (ELK)](https://www.eclipse.org/elk/) via [elkjs](https://github.com/kieler/elkjs), exposed as a single async `layout(graph, options?)` entry point.

- Elements embedded in another element are laid out as a container - the parent is resized (and positioned) by ELK to fit its content, at any nesting depth. Cross-container links are filed under their lowest common ancestor per ELK's hierarchical-edge convention (`elk.hierarchyHandling: INCLUDE_CHILDREN` by default).
- Ports route edges to/from the position JointJS itself already computes for them (`elk.portConstraints: FIXED_POS`); the new `positionPorts` option lets ELK reposition/reorder them instead, writing the computed position back onto the graph.
- Per-cell escape hatches into ELK's option space: `nodeOptions`, `portOptions` and `edgeOptions` callbacks.
- Two examples:
  - `layout-elk-ts` migrated to the new `layout()` entry point (previously hand-rolled ELK export/import).
  - `layout-elk-containers-ports-ts` (new) - a fixed, non-random diagram exercising containers, ports, per-node `positionPorts`, and a width-restricted layout (`elk.aspectRatio` + `elk.layered.wrapping.strategy`).

## Test plan

- [x] `yarn test` in `packages/joint-layout-elk` - 10/10 passing
- [x] `yarn lint` in `packages/joint-layout-elk` - clean
- [x] `tsc --noEmit` for both examples - clean
- [x] Both examples built with webpack and visually verified in a headless browser (containers resize to fit content, ports render with correct labels/anchors, links route correctly within and across containers, `positionPorts` reorders ports on the opted-in nodes only, drawing wraps to stay within the 1000-unit width target)
- [x] `yarn changeset status --since=upstream/master` - reports `@joint/layout-elk -> 0.2.0` (minor) with 4 changesets


examples/layout-elk-containers-ports-ts/src/example.ts
```diff
@@ -0,0 +1,213 @@
...
+// `portsPosition` so ELK orders their ports to minimize crossings (see
+// `index.ts`).
+export const graphJSON: dia.Graph.JSON = {
+    cells: [
+        // Containers
+        {
+            id: 'frontend',
+            type: 'example.Container',
+            attrs: { label: { text: 'Frontend' } },
+            embeds: ['webui', 'mobileui', 'edge']
```
kumilingus, line 20: This is no longer needed (since the `HiearchyIndex` was introduced). `parent` on child is sufficient.

packages/joint-core/src/dia/Link.mjs
```diff
@@ -203,32 +203,47 @@ export const Link = Cell.extend({
     // Labels API
 
     // A convenient way to set labels. Currently set values will be mixined with `value` if used as a setter.
+    // The getter returns a label resolved against `defaultLabel`/the built-in default (see `_getResolvedLabel`) -
+    // the setter always writes the given (raw, unresolved) value, same as before.
     label: function(idx, label, opt) {
```
kumilingus, line 208: This is a breaking change. We need to introduce a new method e.g. getComputedLabel()

packages/joint-core/src/dia/Link.mjs
```diff
@@ -541,18 +556,118 @@ export const Link = Cell.extend({
...
         return label;
+    },
+
+    // A label as given (own `markup`/`attrs`/`size`/`position`, any of which may be missing),
+    // resolved against `defaultLabel` and the built-in default - the same resolution
+    // `LinkView` used to do at render time (see `_mergeLabelAttrs`/`_mergeLabelSize`/
+    // `_mergeLabelPositionProperty` below). Passing `{}` resolves to the pure default.
+    // Any other (custom) property on the label or `defaultLabel` passes through unresolved -
+    // the label's own value wins over `defaultLabel`'s.
+    _getResolvedLabel: function(label) {
```
kumilingus, line 581: Do we need all these private methods? Can we create a new file (`link-labels.js`) that exports a single method that would be used here?

kumilingus (changes requested):

PR #3503, opened 2026-09-12
fix(jumpover): no merge jumps, add isCollinear()

kumilingus:
## Summary

- **`g.Line#isCollinear(line)`** — `true` when all four endpoints lie on one infinite line. Same exact `cross() === 0` test as `containsPoint()` / `equals()`; evaluated in both directions so a zero-length line does not pass trivially (a point off the line → `false`, two points → `true`).
- **Jumpover connector** — replaces the private helpers added in #2255:
  - `overlapExists()` was a *bounding-box* overlap test, so a perpendicular segment that merely touched the link counted as "overlapping" too. In the arrive → run along → depart case it matched the arrival segment first, dropped the harmless collinear segment after it, and left the departure segment in, which drew a spurious jump.
  - `sortPointsAscending()` could undo its own x-sort for non-axis-aligned lines (e.g. `(10,5)-(0,20)`).
  - The filter only ever dropped the segment *following* a collinear one, so the result depended on which way the other link was drawn: a link splitting off a shared stretch produced no jump, the same shape traversed the other way produced a jump at the merge point.

  New rule: for each segment of this link, any segment of another link that is *adjacent to* a segment collinear with it is left out of intersection testing. Such a segment can only touch this link at the shared vertex, where the two links merge or split, never cross it elsewhere. Direction-independent, and handles a link that shares more than one stretch.

  **Behaviour change:** a jump is drawn only where another link genuinely crosses this one. Merge and split points of stacked links never jump. Against `master` this removes the spurious departure jump *and* the jump at the merge point.
- `g.intersection.exists(line, line)` intentionally left unchanged (still `false` for collinear overlap).

Changesets: `minor` for `g.Line`, `patch` for `connectors.jumpover`.

## Test plan

- [x] `karma:geometry` — 353/353; new `isCollinear(line)` module (partial overlap, containment, identical, reversed, touching, disjoint collinear, diagonal, parallel offset, crossing, perpendicular T-junction, shared endpoint only, zero-length on/off line in both call orders). Watched failing first (`isCollinear is not a function`).
- [x] `karma:joint` — 2116/2116 after `shell:rollup-test-bundle` (Karma loads the built `build/test/joint.nodeps.js`, so the bundle must be rebuilt after source changes). Includes the pre-existing `jumpover connector - stacked links do not cause jumps` and a new `jumpover connector - merge and split points` module: merge side of the stacked shape, join → run along → leave in both directions, and a genuine crossing followed by a shared stretch (exactly one jump, at the crossing). All three new tests fail against the previous one-sided filter.
- [x] `grunt test:ts`
- [x] eslint on all touched files
- [x] Six stacked-link scenarios rendered headlessly through the real `jumpover` export on `master` and on this branch (stacked same/reverse direction, arrive → run along → depart in both directions, cross then run along, plain T-junction): jumps only at the genuine crossings on this branch.


MartinKanera:
Expected behaviour:
image

MartinKanera (approved):

